PR #51, opened 2026-08-26
fix(fw-transport): validate YMODEM framing and bound the raw length prefix

seedyjahateh:
## The problem

`core/fw_transport_uart.c` parses data supplied by an untrusted serial peer. The threat model lists this surface explicitly — *"The recovery UART is accessible with enclosure removal"* — and no test covered it. Three defects:

### 1. Out-of-bounds read parsing the YMODEM block-0 header

```c
const char *name = (const char *)block;
size_t name_len = strlen(name);
if (name_len > 0 && name_len < block_size - 1) { ... }
```

`block` is `uint8_t block[YMODEM_BLOCK_SIZE + 2]` holding raw received bytes, with no guarantee of a NUL. A block 0 containing none runs `strlen` off the end of the 1026-byte stack buffer. The bounds guard is applied *after* `strlen` has already read. The decimal-size scan that follows is unbounded too, and can wrap `uint32_t` on a long digit run.

### 2. YMODEM never validates the block number

The XMODEM path checks both the number and its complement:

```c
if (blk_hdr[0] != expected_blk || blk_hdr[1] != (uint8_t)(~expected_blk)) { ... NAK ... }
```

The YMODEM path reads `blk_hdr` and **ignores it**, incrementing `expected_blk` without ever comparing. A block retransmitted after a lost ACK — a routine YMODEM occurrence — is written to flash a second time, shifting every subsequent byte and corrupting the installed image. Out-of-order blocks are accepted too.

### 3. The raw transport trusts its length prefix

The 4-byte little-endian length is peer-supplied and used unchecked, so a declared `0xFFFFFFFF` drives a multi-gigabyte receive loop. To be precise about severity: `eos_fw_update_write()` clamps payload writes to `payload_total`, so this is a stall rather than memory corruption — but the loop has no business starting.

## The approach

- **Block 0 parsing:** scan the filename with an explicit `while (name_len < block_size && block[name_len] != '\0')` bound instead of `strlen`, bound the digit scan to `block_size`, and guard the decimal accumulation against `uint32_t` overflow (treating an overflowing size as unknown).
- **Sequencing:** validate `blk_hdr[1] == (uint8_t)~blk_hdr[0]`, require block 0 first, and accept only `expected_blk`. An exact retransmission of the previous block is ACK'd but **not** written again; anything else is NAK'd.
- **Raw length:** reject zero or anything larger than `ctx->slot_size` before reading any payload, replying NAK with `EOS_ERR_FULL`.

Requiring the first block to be block 0 is a tightening — the previous code treated whatever arrived first as the header. Standard YMODEM senders always transmit block 0, and the old behaviour would have misparsed a data block as a header anyway.

The protocol comment blocks at the top of each transport section are updated to document the new constraints; these are the only description of the wire format in the repo.

## Testing

New `tests/unit/test_fw_transport.c` (10 cases), registered in `tests/CMakeLists.txt` alongside the existing suites. It drives both transports through their **public ops tables** (`eos_fw_transport_uart_ymodem()` / `_raw()`) using a scripted UART and simulated flash installed via `eos_hal_init()` — the same mocking idiom as `test_bootctl.c`, so no statics needed exposing.

Confirmed failing against unmodified `fw_transport_uart.c`:

```
test_ymodem_valid_transfer_is_written              [OK]
test_ymodem_duplicate_block_is_not_written_twice   [FAIL] run_ymodem() == EOS_OK
```
```
test_raw_valid_transfer_is_written                 [OK]
test_raw_oversized_length_is_rejected              [FAIL] run_raw() == EOS_ERR_FULL
```
```
test_ymodem_first_block_must_be_zero               [FAIL] tx_contains(XM_NAK)
```

The two `valid_transfer` cases pass before *and* after — they are regression guards proving the added validation does not reject well-formed transfers.

Verification (GCC 16.2.0, `-DEBLDR_BUILD_TESTS=ON`):

| Check | Result |
|---|---|
| `test_fw_transport` | PASS, 10/10 |
| `ctest` full suite | 14/15 |
| `gcc -std=c11 -Wall -Wextra -Wpedantic` on the changed file | zero warnings |

## Considerations and limitations

**One pre-existing failure, untouched.** `test_recovery` does not link on `master`: `core/recovery.c` calls `eos_boot_log_append` / `_read` / `_get_head`, but `boot_log.c` is compiled into `eboot_stage1`, not `eboot_core`. Baseline before my change was 13/14; it is 14/15 after, i.e. unchanged. This is already the subject of #35, #39, #41 and #49, so I have deliberately not duplicated it — but it does mean this branch cannot show a fully green `ctest` until one of those lands.

**The out-of-bounds read is not covered by a failing-before test.** Defects 2 and 3 have deterministic regression tests, as shown above. Defect 1 does not: with the old code the over-read lands in adjacent stack whose contents are not deterministic, so the observable outcome is usually identical. It needs ASan or Valgrind to demonstrate, and `EBLDR_SANITIZE=ON` is not usable with the MinGW toolchain I validated on (no ASan runtime). `test_ymodem_header_without_nul_is_bounded` pins the correct behaviour and the new scan is provably in-range by construction, but I want to be straight that it would not have caught the original bug on its own. Running the suite under ASan on a glibc host would close that gap.

**Environment.** Validated on Windows with GCC 16.2.0 via a portable MinGW-w64 toolchain; the MSVC and Clang CI legs were not run locally.


seedyjahateh:
Thanks — all three points were valid and are addressed in `3f6ab20`.

**Valgrind registration.** `test_fw_transport` was the only unit test missing from the loop in `tests/CMakeLists.txt`; it is now included. This one mattered more than it looks: the PR description stated that the block-0 over-read could not be demonstrated without a memory checker, and that loop is exactly the regime that would catch it.

**STX coverage.** Correct that the 1024-byte path was untested, and it turned out to be the more important of the two framings. `ymodem_receive()` declares `uint8_t block[YMODEM_BLOCK_SIZE + 2]` — 1026 bytes — and **only a full STX block fills every one of them**. With a 128-byte SOH block most of the buffer is untouched, so the old `strlen()` stayed in bounds and merely read indeterminate values. Two tests added; the fill byte is `0x21` because its CRC over 1024 bytes is `0xE940`, so neither CRC byte is NUL either and nothing stopped the old scan before the end of the buffer.

`test_ymodem_stx_duplicate_block_is_not_written_twice` fails against the unmodified transport, confirming the sequencing fix is exercised on the STX path and not just SOH.

**Overflow-safe simulator bounds.** Fixed via a `sim_range_ok()` helper. The original form was copied from the existing `test_bootctl.c` idiom, but you're right that a simulator which fails open converts a bounds bug in the code under test into heap corruption instead of a legible failure.

---

**One correction to the PR description, in the PR's favour.** It said the over-read was verified by inspection only. I have since demonstrated it directly: a standalone harness places a 1026-byte block flush against a `PAGE_NOACCESS` guard page, fills it with `0x21` so no NUL exists anywhere, and runs the original `strlen()` line. It faults with `0xC0000005` at the guard page:

```
block spans [000001eefa0f0bfe, 000001eefa0f1000), guard page at 000001eefa0f1000
calling strlen() exactly as the original code did...
exit: -1073741819
```

That harness is diagnostic scaffolding and is not part of this PR — the shipped equivalent is the STX test now wired into the Valgrind loop. Verification remains GCC 16.2.0 on Windows; Valgrind itself is unavailable there, so the Valgrind leg is **NOT RUN** locally.

Full suite after the change: `test_fw_transport` 12/12, `ctest` 14/15 — the outstanding failure is the pre-existing `test_recovery` link error on `master`, unchanged.

srpatcha (dismissed):

PR #55, opened 2026-08-27
fix: restore eboot_core buildability broken by unresolved merge conflicts

tejasmhadgut:
## Issue

`master` currently fails to build `eboot_core`. Bisecting the recent merge history
turned up **five independent, pre-existing regressions**, all traceable to
conflict-resolution mistakes in earlier merged PRs — not something introduced by
this branch.

1. **`core/recovery.c`** — `recovery_handle_write()` declares `slot_size` twice in
   the same scope. Introduced when PR #33 replaced a working call to
   `eos_recovery_write_in_range()` with an inline reimplementation, without noticing
   `slot_size` was already declared earlier in the function.
   ```
   error: redefinition of 'slot_size'
   ```

2. **`include/eos_image.h`** — `eos_crc32()`'s declaration didn't match its own
   definition or doc comment (wrong return type, wrong parameter count). A stale
   prototype, unrelated to any merge.

3. **`core/ed25519_verify.c`** — PR #46's merge commit (`19c7e72`) resolved a
   conflict between two independently-written Ed25519 implementations by mixing
   incompatible halves: it kept the setup from `master`'s `gf`-based implementation
   but pasted in the PR branch's `ge25519_p3`-based comparison logic, calling an
   undefined function (`sc_reduce`) and leaving `diff` undefined at the return
   statement. This doesn't compile — but if it had (e.g. via a naive `diff = 0`
   patch), it would have silently accepted every signature.

4. **`core/sha512.c` + `CMakeLists.txt`** — the same PR #46 replaced a working
   `eos_sha512_init/update/final` implementation (the API `eos_crypto_boot.h`
   declares and `ed25519_verify.c` depends on) with an incompatible,
   differently-named implementation (`sha512_init`, etc., under a new
   `eos_sha512.h`). The subsequent merge-conflict resolution then dropped
   `core/sha512.c` and `core/rollback.c` from `eboot_core`'s source list entirely,
   and left `core/boot_log.c` listed twice.

5. **`include/eos_boot_log.h`** — `eos_boot_log_read()`'s declaration was stale
   (unmodified since v0.1.0) and didn't match the real implementation in
   `core/boot_log.c` or its caller in `core/recovery.c`, both of which use a
   single-index read.

## Approach

- **`recovery.c`**: removed the duplicate declaration.
- **`eos_image.h`**: corrected `eos_crc32()`'s prototype to match its definition
  and doc comment.
- **`ed25519_verify.c`**: restored the pre-conflict verification tail
  (`scalarmult`/`scalarbase`/`point_add`/`point_pack` + byte-compare into `diff`),
  recovered from commit `02b7dac` (`master` immediately before the bad merge). All
  of that code's helper functions were still present in the file, just orphaned
  since the conflict — nothing new was written, only restored.
- **`sha512.c`**: restored from the same pre-conflict commit, confirmed to
  implement the API actually depended upon elsewhere in the codebase. Removed the
  now-fully-unused `eos_sha512.h`.
- **`CMakeLists.txt`**: added `core/sha512.c` and `core/rollback.c` back to
  `eboot_core`'s source list; removed the duplicate `core/boot_log.c` entry.
- **`eos_boot_log.h`**: corrected the declaration and doc comment to match the
  real single-index-read implementation.

Every restoration above is evidence-backed against actual pre-conflict commits
(`02b7dac`), not a guess at intended behavior.

## Testing

```
$ cmake --build build --target eboot_core test_recovery test_image_verify test_ed25519 --parallel
...
[100%] Built target test_recovery
[100%] Built target test_image_verify
[100%] Built target test_ed25519

$ ctest --test-dir build --output-on-failure -R "test_recovery|test_image_verify|test_ed25519"
100% tests passed, out of 3
```

`test_ed25519` is the load-bearing check here — it exercises the restored
point-addition/comparison logic end-to-end (not just that it compiles), and passing
confirms the restored implementation is cryptographically correct, not merely
syntactically valid.

`git status` is clean against the commit; the working tree used for testing exactly
matches what's in this PR.

## Limitations

- This PR does **not** attempt a full `ctest --test-dir build` run across every
  target — two pre-existing, unrelated test files are still broken on `master` and
  are out of scope here (filed as a separate issue: #56):
  - `tests/unit/test_slot_manager.c` — ~20 test functions are called from `main()`
    but never defined, from PR #37's incompletely-applied patch.
  - `tests/unit/test_boot_log.c` — defines a local mock of `eos_boot_log_read()`
    that conflicts with the real implementation once linked against `eboot_core`
    (a duplicate-symbol issue needing a real redesign, not a one-line fix).
- I did not audit every other file in `core/` for the same class of bad-merge
  damage — I fixed what was needed to get `eboot_core` linking and the directly
  affected tests passing. There may be other, undiscovered instances of the same
  pattern elsewhere in the tree.


srpatcha:
### Review — verified, and this should merge first

**`eBoot` `master` does not currently compile.** Built it to confirm:

```
core/recovery.c:287:14: error: redefinition of 'slot_size'
core/image_verify.c:63:10: error: conflicting types for 'eos_crc32';
    have 'uint32_t(uint32_t, size_t)'
```

Built this branch merged: **configure and build both succeed, exit 0.**

So this is not an improvement to a working tree — it repairs a repository whose default branch cannot be built at all. That makes it the highest-priority open PR in this repo by some distance.

Two things worth raising alongside it:

1. **`ci.yml` here already triggers on `master` pull requests**, unlike the eos one. So CI either was failing and the breakage merged anyway, or the job does not compile `eboot_core`. Either answer is worth chasing — a red default branch that nobody notices is how the next one lands too.

2. The `sha512.c` change is large (338 lines). Given that ADR-011 (in the eos repo) records that `eBoot`'s Ed25519 is **not RFC 8032** — it hashes with SHA-256 where the standard requires SHA-512, so both RFC 8032 §7.1 vectors are rejected and signatures from the project's own signing tool do not verify — it would help to say in the PR description whether this touches that path or is purely a build repair. If it is purely a build repair, say so explicitly; if it changes hashing behaviour, that interacts with an open architectural decision.

Nothing here needs to change for it to merge. Please get this in.

tejasmhadgut:
Thanks for the review — chased down both.

**CI.** Confirmed: `CI — eBoot`, `eBoot Build & Test`, and `CodeQL` are all `failure` on `master`'s recent pushes, so the breakage did merge past red CI. There's a second, related piece: this PR's own checks are sitting in `action_required`, not pass/fail — `gh api repos/embeddedos-org/eBoot/pulls/55` shows `mergeable_state: "blocked"`. Actions requires a maintainer to manually approve a run for outside/fork contributors before it executes at all. That's plausibly the actual mechanism behind "nobody notices" — a first-time PR's CI signal doesn't exist until someone clicks approve, so the gate that would normally catch this on the way in isn't live by default.

**ADR-011 / SHA-512.** I don't have access to ADR-011 itself — it's on `reconcile/tier-1`, which isn't pushed to any remote I can reach, so I can't speak to what it records directly. What I can say from reading the code:

`eos_ed25519_verify()` (the function this PR restores) hashes with real SHA-512 — `core/sha512.c` has the actual FIPS 180-4 round constants (`0x428a2f98d728ae22ULL`, etc.), correct 64-byte digest / 128-byte block sizes. Not SHA-256 mislabeled. `test_ed25519_rfc8032_vectors_accepted` feeds real RFC 8032 §7.1 vectors straight into this function and passes.

The SHA-256 pattern is real, but it's one layer up: `core/crypto_boot.c`'s `eos_crypto_verify_signature()` wrapper pre-hashes the message with SHA-256 before calling into `eos_ed25519_verify()` — its own doc comment calls this out explicitly as an mcuboot-style hash-then-verify convention. That file is untouched by this PR (`git diff master..HEAD --stat -- core/crypto_boot.c` is empty).

So: purely a build repair with respect to this question. The primitive this PR restores is independently RFC-8032-conformant by the test above; the pre-hash-vs-raw-message design choice ADR-011 is presumably about lives in a different function this PR never touches.

**#57 / #58.** Both are fixing overlapping ground independently — worth saying so given the CI discussion above, since three people converging on the same red `master` within a day is itself evidence for point 1. #58 already proposed rebasing onto whichever of #55/#57 lands first and keeping just the test-integrity half (`test_slot_manager.c`, `test_boot_log.c`, the board-dispatch duplication, the ARM CI toolchain path) — happy to have that land on top of this once it's in.

srpatcha (commented):
Correct diagnosis, and the earliest of the four PRs that independently found this
breakage. It is narrower than #58, which is the one already approved, so I want
to be specific about what the gap is rather than just saying "duplicate".

## Every file here is also in #58

```
CMakeLists.txt
core/ed25519_verify.c
core/recovery.c
core/sha512.c
include/eos_boot_log.h
include/eos_image.h
```

Six files, zero unique to this PR.

## What it leaves behind

Built standalone:

```
60 build errors
81% tests passed, 3 tests failed out of 16

test_slot_manager.c:39:5: error: 'erase_result' undeclared
test_slot_manager.c:40:5: error: 'erased_addr' undeclared
test_slot_manager.c:46:36: error: 'SLOT_A_ADDR' undeclared

The following tests FAILED:
  11 - test_slot_manager (Not Run)
  12 - test_boot_log (Not Run)
```

`eboot_core` builds — the title's claim holds — but `test_slot_manager.c` and
`test_boot_log.c` are broken *independently* of it, by the same merge, and are
not repaired here. Those two are tracked as #56. #58 fixes them, which is why it
reaches 16/16.

Nothing wrong with the scope you chose; the merge left several separable
breakages and you fixed one. It just means this cannot land on its own without
leaving the suite red.

## Recommendation

Close in favour of #58. If any hunk here resolves something differently and
better than #58 does, point at it and I will fold it across with attribution —
that is what happened with #93's and #94's comment improvements in eos, and with
#94's `sync.c` wording.

There is one thing in this area genuinely worth picking up, if you want it:
`core/ed25519_verify.c` on **both** this branch and #58 accepts a forged
signature over any message when the public key is a low-order point:

```
identity_pub[32] = {1}   /* identity point */
identity_sig[64] = {1}   /* R = identity, S = 0 */
eos_ed25519_verify(...)  ->  rc=0, ACCEPTED
```

A complete secure-boot bypass, filed as #73. #57 fixes it with a subgroup check.
Neither this PR nor #58 has one — worth knowing, since the build repair is the
part everyone converged on and the crypto is the part that actually matters.

srpatcha (commented):
## Review — eBoot#55 "fix: restore eboot_core buildability broken by unresolved merge conflicts"
head: 9fbfc919e6fa8d9104f24aa6d0ad4bf53ce9bf5c  author: tejasmhadgut  ci: none in bundle

**Verdict:** The work was right and the diagnosis was right. It has since been overtaken
— all five defects are fixed on `master`, so merging this now would conflict in five
files and revert later work rather than repair anything. This should be closed, not
rebased.

### Findings
|---|---|---|---|---|
| 1 | High | (whole PR) | **Premise no longer holds.** The PR exists because `master` did not build. It does now, and each of the five named defects is independently fixed upstream — see the table below. Merging would resolve five conflicts in favour of a six-day-old tree. | Close with credit. The repair landed; the branch is the wrong vehicle for what is left. |
| 2 | Medium | `core/sha512.c` | `git merge-tree` conflicts here. `master`'s file is 252 lines with explicit `load_be64()` / `store_be64()` helpers (`core/sha512.c:98`, `:110`); this branch's is 155 lines. Both are byte-order-explicit and both look correct, but resolving toward this branch drops a later independent rewrite for no functional gain. | If any part of this branch is salvaged, do not take `sha512.c`. |
| 3 | Medium | `CMakeLists.txt`, `core/ed25519_verify.c`, `core/recovery.c`, `include/eos_boot_log.h` | The other four conflicts. `CMakeLists.txt` alone differs by 160 lines against `master`. The static function set in `ed25519_verify.c` is now identical between the two, so there is nothing left to recover there. | Same — no salvage needed. |

### Status of each of the five defects on current `master` (`13a7a02`)
| # | Defect as filed | State on `master` | Evidence |
|---|---|---|---|
| 1 | `recovery.c` redefines `slot_size` | Fixed | `core/recovery.c:281` and `:317` are separate functions; one declaration each |
| 2 | `eos_crc32()` prototype mismatch | Fixed | `include/eos_image.h:149` `uint32_t eos_crc32(uint32_t addr, size_t len)`, plus a new `eos_crc32_checked()` at `:135` |
| 3 | `ed25519_verify.c` mixed halves, undefined `sc_reduce`/`diff` | Fixed | no `sc_reduce`; `diff` declared at `core/ed25519_verify.c:143` and `:442` |
| 4 | `sha512.c`/`rollback.c` dropped from `eboot_core`, `boot_log.c` listed twice | Fixed | both present in the `eboot_core` list; `boot_log.c` appears once |
| 5 | `eos_boot_log_read()` stale declaration | Fixed | `include/eos_boot_log.h:72` is the single-index form |

`master` reached this through `b0b2ab9` ("repair three merge regressions left by today's
PR flurry"), `038f624`, `b7e07d4` and others. Two items from the PR's own *Limitations*
section are also resolved: `test_slot_manager` and `test_boot_log` are both registered
and running on `master` (20 `add_test` entries in `tests/CMakeLists.txt`).

The CMake source list is the one place where something from this branch is still owed —
`master` still registers `core/sha512.c` and `core/rollback.c` **twice** and omits
`core/fdt_loader.c` entirely. That is #57's and #84's territory now, not this one's.

### Architecture conformance
Conformed. §21 Tier 1 (`eBoot`, Foundation); a build repair inside the owning repo, no
dependency edges added, §5.1 untouched. The ADR-011 question raised in review was
answered correctly in-thread: the restoration touched `eos_ed25519_verify()`, which
hashes with real SHA-512, and not `core/crypto_boot.c`'s SHA-256 pre-hash wrapper —
`core/crypto_boot.c` is still untouched by this branch.

### Proposed changes
Close. If the maintainers would rather not lose the branch, the only content still
missing from `master` is the two duplicate CMake entries, and #57 already carries that
change with a regression test attached — reviewed there.

### Not checked
- **I did not build anything.** The brief forbids checking out or otherwise disturbing
  the repos, and `eBoot`'s working tree is currently on `fix/ed25519-low-order-keys`.
  Every statement above is from `git show` / `git merge-tree` against `origin/master`
  and `refs/pull/55/head`. The claim "`master` builds now" is *inferred* from the five
  defects being individually absent, not verified by a compile.
- `checks.txt` is empty for this bundle; I have no CI signal for this head.
- The PR's *Limitations* note about undiscovered instances of the same bad-merge pattern
  elsewhere in `core/` — I did not audit for those either.


---
*Automated architecture review of `9fbfc919e6fa` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #63, opened 2026-08-29
Harden recovery VERIFY against slot boundary violations

swayam-2003:
## Summary
- **WRITE handler:** Removed a duplicate `slot_size` declaration that broke compilation; routed bounds checking through the existing `eos_recovery_write_in_range()` helper.
- **VERIFY handler:** Added the same slot-capacity guard used in `slot_manager.c` so UART recovery cannot stream integrity reads past a slot when `image_size` exceeds real slot capacity.

## Approach
Reused the proven bounds check from boot-time slot verification (PR #50). Extended `tests/unit/test_recovery.c` with a VERIFY regression test.

## Testing
```bash
cmake -B build -DEBLDR_BUILD_TESTS=ON
ctest --test-dir build -R test_recovery --output-on-failure
```

## DCO
Signed-off-by: Swayam Nayak <154440440+swayam-2003@users.noreply.github.com>


srpatcha (approved):
## Verified — the fix is right, and the conflict with #58 is one dropped line

`offset` and `len` arrive on the recovery UART and reach `eos_hal_flash_write(base + offset, ...)`. Without a bound, a recovery client can write past the slot into the other slot, the boot-control blocks, or the boot log — from a serial port, on a device that is by definition already in trouble.

Reusing `eos_recovery_write_in_range` is the right call. That helper already exists on `master` and gets the arithmetic right:

```c
/* Check len first so slot_size - len cannot underflow. */
if ((uint32_t)len > slot_size || offset > slot_size - (uint32_t)len)
```

Checking `len` before subtracting is the part that matters — `offset + len > slot_size` on 32-bit values is exactly the wrap that eos#87 is fixing in the OTA path this week.

## The conflict

This does not build merged with #58, but the cause is mechanical, not a disagreement:

```
core/recovery.c:286: error: 'slot_size' undeclared
```

Both branches declare `uint32_t slot_size = eos_hal_slot_size(slot);` — yours at line 278, #58's at 286 — and both add a bounds check. The merge kept **your call** and **#58's declaration line**, then #58's own inline check went away with its declaration. One line is missing, not a design clash.

Resolving it is restoring that declaration:

```c
uint32_t base = eos_hal_slot_addr(slot);
uint32_t slot_size = eos_hal_slot_size(slot);
```

Done that locally on top of #58: **0 build errors, `ctest` 16/16**.

## Which check survives

#58 inlines `(uint64_t)offset + len > (uint64_t)slot_size`; yours calls the existing helper. Both are correct. I would keep yours — the helper is already in the tree, it is already tested, and a second inline copy of the same bound is how the two drift. Worth saying so on #58 rather than deciding silently, so I have.

## Blocker outside this PR

eBoot sets `required_signatures: true` on `master` and commits here are unsigned, as are every contributor's. Nothing in this repository is mergeable until that policy changes — I have raised it with the maintainer. Also note `master` does not build at all right now; #58 repairs it and needs to land first.

swayam-2003:
Rebased onto latest upstream/master (ae1e0f7). Previous CI failures were from a stale base with upstream merge compile regressions (now fixed in #69). The WRITE bounds path is covered upstream (#74); this PR now adds only the VERIFY slot-capacity check before integrity reads + unit test.

PR #64, opened 2026-08-29
fix(recovery): reject out-of-range UART writes via the existing helper

ShrenikMensinkai:
## Summary

UART recovery WRITE takes offset and length from the serial packet. The write handler declared slot size twice (invalid C) and did not call `eos_recovery_write_in_range`, so wrap and zero-base checks were skipped. This PR uses that helper once, exposes it in a public header, and adds host tests for those cases.

This addresses:

- **Issue:** merge leftover in the write handler; wire-controlled offset could miss range checks
- **Approach:** single slot-size variable; NACK unless the existing helper accepts the write
- **Testing:** new helper unit tests plus the existing out-of-bounds WRITE protocol test
- **Limits:** WRITE path only; VERIFY still has no slot-capacity check; no hardware run on this machine

## Type of Change

- [ ] feat — New feature
- [x] fix — Bug fix
- [x] test — Add or fix tests
- [ ] build — Build system or dependency changes

## Changes

- Removed the duplicate slot-size declaration in the recovery write handler
- Route WRITE through `eos_recovery_write_in_range` before receiving payload or programming flash
- Added `include/eos_recovery.h` so tests can call the helper and enter recovery
- Added host tests for zero base, zero length, past-slot offset, base+offset wrap, and a valid in-slot write

## Testing

- [ ] Unit tests pass (`ctest --test-dir build --output-on-failure`)
- [ ] Integration tests pass

CMake and a C compiler were not available on the Windows host used for this change. Please run a native build with `EBLDR_BUILD_TESTS=ON` and `ctest -R test_recovery` before merge.

## Pre-Submission Checklist

- [ ] Code compiles without warnings (`-Wall -Wextra -Werror` for C)
- [ ] All existing tests pass
- [x] Documentation updated if API changed (new public header)

## Related Issues

None. Found while reviewing the recovery write path after a recent pull.

## Screenshots / Logs

Not applicable. Host unit tests only; no board or UART capture.

## Additional Notes

- Recovery VERIFY still does not apply the slot-capacity check that the slot manager already has. That can be a follow-up.
- Recovery authentication and RESET-without-auth are unchanged.
- Tests use simulated flash and a scripted UART, not a real probe.
- The CRC fail-open path is already fixed upstream and is not part of this PR.


srpatcha (dismissed):
## Verified — and the header is the part I most want

You and #63 both route the recovery write through `eos_recovery_write_in_range`, and both are right to: the helper already exists, already gets the arithmetic right, and a second inline copy of the same bound is how the two drift.

What this PR has that #63 does not is **`include/eos_recovery.h`**. On `master` that file does not exist:

```
$ git show origin/master:include/eos_recovery.h
fatal: path 'include/eos_recovery.h' does not exist in 'origin/master'
```

So `eos_recovery_write_in_range` is a function with no declaration anywhere — reachable only by whoever already knows it is in `recovery.c`. That is the same shape as `eos_hal_linux_register()` in eos, which implements the entire host HAL backend and is called from nowhere because nothing declares it. A function nobody can find is a function nobody uses, and the second person to need this bound writes their own instead. Which is exactly what happened: `master` carries the helper *and* an inline copy of the same check.

Replacing the inline `(uint64_t)offset + len > (uint64_t)slot_size` with the helper is also a small correctness gain, not just tidying — the helper checks `len` before subtracting, so `slot_size - len` cannot underflow, and it rejects a zero base, which the inline version did not.

## Overlap and merge order

This conflicts with #58, and so does #63, for the same mechanical reason: both branches declare `uint32_t slot_size = eos_hal_slot_size(slot);` at different lines, and the merge keeps one side's call with the other side's declaration. The build then fails with

```
core/recovery.c:286: error: 'slot_size' undeclared
```

Restoring that one line fixes it. Verified on top of #58: **0 build errors, `ctest` 16/16**.

Between this and #63 I would take **this one**, because of the header — and then the delta from #63 is small enough to fold in. Worth the two of you agreeing rather than both rebasing; I have said the same on #63.

## Blocker outside this PR

eBoot sets `required_signatures: true` on `master` and commits here are unsigned, as are every contributor's — nothing in this repository is mergeable until that policy changes. And `master` does not build at all; #58 repairs it and must land first.

PR #82, opened 2026-09-01
fix(secure-boot): a debug lock that failed must not report a successful boot

Kartikey1306:
`cfg.lock_debug` asks for SWD/JTAG to be closed before the verified image runs. Step 7 of `eos_secure_boot()` honoured that request like this:

```c
if (cfg->lock_debug) {
    eos_secure_boot_lock_debug();
}

/* ---- Step 8: Record successful attestation ---- */
attest_record(2, hdr.image_version, hdr.hash, NULL, EOS_SBOOT_OK);
return EOS_SBOOT_OK;
```

`eos_secure_boot_lock_debug()` returned **void** and discarded the result of the OTP write that actually blows the fuse. When that write failed, boot continued, attestation recorded `EOS_SBOOT_OK`, and the device ran the image **with its debug port open** — the exact condition the policy exists to prevent, reported as a clean secure boot.

## The interesting case is not a flaky fuse

`eos_hal_otp_write()` returns `EOS_ERR_NOT_SUPPORTED` when the board provides no `otp_write` hook at all. So on every such board, **`lock_debug: true` was silently a no-op** — that is the default configuration, not an edge case.

## The change

`eos_secure_boot_lock_debug()` returns `int`, and a caller that asked for the lock and did not get it now fails with **`EOS_SBOOT_ERR_POLICY`** — a code that already existed for exactly this ("Boot policy violation") — with the failure recorded in the attestation log instead of a success.

## Why this was never observable

**`core/secure_boot.c` is not in `CMakeLists.txt`.** The module has never been compiled, so this path could not run and could not be tested:

```
$ grep -c "core/secure_boot.c" CMakeLists.txt
0
```

Added the one line that builds it — **the same line #72 adds, written identically**, so whichever lands first leaves the other a trivial rebase.

## Tests

`tests/unit/test_secure_boot_policy.c` covers the three outcomes: fuse written, write failing, and a board with no `otp_write`. Kept in its own file so it does not collide with the `test_secure_boot.c` that #72 introduces.

Against `master` the new test **does not compile**:

```
error: invalid operands to binary expression ('void' and 'int')
```

because there is no result to check. That is the defect stated as a compile error rather than a runtime one.

## Verification

| | |
|---|---|
| build | clean, 0 errors |
| `ctest --no-tests=error` | **20/20 pass** |
| `pytest tests/` | 30 passed |

**Stacked on #77** — master's test suite does not compile without it (`test_image_verify.c` lost a line continuation in #70).

## Relationship to #72

Complementary, not competing. #72 builds the module and stops it booting plaintext; this fixes a separate fail-open inside the same function's policy handling. The only textual overlap is the one-line `CMakeLists.txt` addition, deliberately identical.

### DIFF
--- a/core/ed25519_verify.c
+++ b/core/ed25519_verify.c
@@ -289,6 +289,16 @@ static int point_is_identity(gf p[4])
     return diff == 0;
 }
 
+static void scalarbase(gf r[4], const uint8_t *s)
+{
+    gf q[4];
+    fe_copy16(q[0], BX);
+    fe_copy16(q[1], BY);
+    fe_copy16(q[2], gf1);
+    fe_mul(q[3], BX, BY);
+    scalarmult(r, q, s);
+}
+
 /* Reject a public key outside the prime-order subgroup.
  *
  * Decoding a point is not enough. Ed25519 has eight points of low order, and
@@ -304,51 +314,9 @@ static int point_is_identity(gf p[4])
  * so there is no separate constant to transcribe wrongly: a mistyped L would
  * reject valid keys, and only in the field.
  *
- * A arrives negated from unpackneg(). [L](-A) = -[L]A and the identity is its
- * own negation, so neither condition is affected by the sign.
- *
- * Formulation taken from eBoot#57 by @muhammadburhandevv-hub, which reached
- * this before I did and states both conditions in one expression.
+ * The key arrives negated from unpackneg(). [L](-A) = -[L]A and the identity
+ * is its own negation, so neither condition is affected by the sign.
  */
-static int key_has_prime_order(gf A[4])
-{
-    uint8_t order_l[32];
-    gf q[4], multiple[4];
-    int i;
-
-    for (i = 0; i < 32; i++)
-        order_l[i] = (uint8_t)ORDER_L[i];
-    for (i = 0; i < 4; i++)
-        fe_copy16(q[i], A[i]);
-
-    scalarmult(multiple, q, order_l);
-    return point_is_identity(multiple) && !point_is_identity(A);
-}
-
-static void scalarbase(gf r[4], const uint8_t *s)
-{
-    gf q[4];
-    fe_copy16(q[0], BX);
-    fe_copy16(q[1], BY);
-    fe_copy16(q[2], gf1);
-    fe_mul(q[3], BX, BY);
-    scalarmult(r, q, s);
-}
-
-static int point_is_identity(gf p[4])
-{
-    uint8_t encoded[32];
-    point_pack(encoded, p);
-
-    uint8_t diff = (uint8_t)(encoded[0] ^ 1U);
-    for (int i = 1; i < 32; i++)
-        diff |= encoded[i];
-    return diff == 0;
-}
-
-/* Public keys must be non-identity points in Ed25519's prime-order subgroup.
- * Merely decoding a point is insufficient: an identity or torsion key can
- * make the verification equation true without knowledge of a private key. */
 static int public_key_is_valid_subgroup(gf public_key[4])
 {
     uint8_t order_l[32];
--- a/core/secure_boot.c
+++ b/core/secure_boot.c
@@ -182,8 +182,19 @@ eos_secure_boot_result_t eos_secure_boot(const eos_secure_boot_config_t *cfg,
     }
 
     /* ---- Step 7: Lock debug interfaces ---- */
+    /*
+     * A policy that asked for the debug port to be closed and did not get
+     * it is a policy violation, not a detail. The result used to be
+     * discarded, so a board whose OTP write failed -- or one with no
+     * otp_write at all, where the HAL returns EOS_ERR_NOT_SUPPORTED --
+     * booted with SWD/JTAG open while attestation recorded EOS_SBOOT_OK.
+     */
     if (cfg->lock_debug) {
-        eos_secure_boot_lock_debug();
+        if (eos_secure_boot_lock_debug() != EOS_OK) {
+            attest_record(2, hdr.image_version, hdr.hash, NULL,
+                          EOS_SBOOT_ERR_POLICY);
+            return EOS_SBOOT_ERR_POLICY;
+        }
     }
 
     /* ---- Step 8: Record successful attestation ---- */
@@ -212,17 +223,21 @@ int eos_secure_boot_verify_key(const uint8_t key_hash[32])
     return secure_compare(key_hash, otp_hash, 32);
 }
 
-void eos_secure_boot_lock_debug(void)
+int eos_secure_boot_lock_debug(void)
 {
     /* Write lock pattern to eFuse debug lock register */
     uint8_t lock = 0xFF;
-    eos_hal_otp_write(OTP_DEBUG_LOCK_OFFSET, &lock, 1);
+    int rc = eos_hal_otp_write(OTP_DEBUG_LOCK_OFFSET, &lock, 1);
+    if (rc != EOS_OK)
+        return rc;
 
     /* On Cortex-M: disable DAP access via DHCSR if supported */
 #if defined(__ARM_ARCH)
     /* Some MCUs support disabling debug via DBGMCU register */
     /* *((volatile uint32_t *)0xE0042004) = 0; */
 #endif
+
+    return EOS_OK;
 }
 
 int eos_secure_boot_update_rollback(uint32_t new_version)
--- a/core/secure_boot.c
+++ b/core/secure_boot.c
@@ -102,18 +102,41 @@ eos_secure_boot_result_t eos_secure_boot(const eos_secure_boot_config_t *cfg,
         }
 
         /* ---- Step 4: Verify signing key against OTP root-of-trust ---- */
-        /* Extract key hash from TLV area */
-        /* The TLV_KEYHASH entry contains SHA-256 of the public key */
+        /*
+         * This step is PLANNED, not implemented: comparing the image's
+         * TLV_KEYHASH against the OTP anchor needs TLV extraction that does
+         * not exist here yet. What it must not do in the meantime is report
+         * success.
+         *
+         * It used to. An otp_read failure was discarded and boot continued,
+         * and when the read succeeded and the anchor was provisioned the body
+         * of the `if` was a comment -- so a device whose root of trust *is*
+         * provisioned booted an image signed by any key the image carried,
+         * and step 8 recorded EOS_SBOOT_OK. That is the same fail-open this
+         * PR removes from step 7, two steps earlier.
+         *
+         * Until the comparison exists, a provisioned device fails closed. An
+         * unprovisioned one (all-zero anchor) is unchanged: there is nothing
+         * to check against, and refusing would brick every un-provisioned
+         * board.
+         */
         uint8_t otp_key_hash[32];
         rc = eos_hal_otp_read(OTP_KEY_HASH_OFFSET, otp_key_hash, OTP_KEY_HASH_SIZE);
-        if (rc == EOS_OK) {
-            /* Check if OTP key hash is provisioned (not all-zeros) */
-            uint8_t zeros[32] = {0};
-            if (secure_compare(otp_key_hash, zeros, 32) != 0) {
-                /* OTP is provisioned — must match */
-                /* In a full implementation, extract key hash from TLV and compare */
-                /* For now, the signature verification implicitly uses the embedded key */
-            }
+        if (rc != EOS_OK) {
+            /* The anchor could not be read, so it cannot be checked. A
+             * verification step that cannot run must fail, not pass. */
+            attest_record(2, hdr.image_version, hdr.hash, NULL,
+                          EOS_SBOOT_ERR_SIGNATURE);
+            return EOS_SBOOT_ERR_SIGNATURE;
+        }
+
+        uint8_t zeros[32] = {0};
+        if (secure_compare(otp_key_hash, zeros, 32) != 0) {
+            /* Provisioned, and nothing here compares against it. Refuse
+             * rather than boot on a key this function has not checked. */
+            attest_record(2, hdr.image_version, hdr.hash, NULL,
+                          EOS_SBOOT_ERR_SIGNATURE);
+            return EOS_SBOOT_ERR_SIGNATURE;
         }
     }
 
--- a/include/eos_image.h
+++ b/include/eos_image.h
@@ -108,7 +108,8 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, tlv_hash) +
 
 /* Every remaining field, pinned.
  *
- * Four of the fourteen fields were asserted. Transposing two adjacent
+ * Three of the thirteen field offsets were asserted (the fourth pre-existing
+ * assert is sizeof, which is not a field). Transposing two adjacent
  * same-width fields moves neither sizeof nor any of those four offsets, so it
  * compiled clean: with load_addr and entry_addr swapped, all four existing
  * asserts still passed and the bootloader would load an image at its entry
@@ -132,15 +133,18 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, flags) == 24,
                       "flags must stay at offset 24");
 EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, sig_len) == 61,
                       "sig_len must stay at offset 61");
-EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, reserved) == 62,
-                      "reserved[] must stay at offset 62");
+/* tlv_len and tlv_hash are asserted above, where #93 introduced them; the
+ * 30 bytes they occupy are the ones this block used to pin as reserved[]. */
 
 /* Field widths. An offset assert cannot see a field growing into padding that
- * happens to keep every later offset -- reserved[] absorbs exactly that. */
+ * happens to keep every later offset -- the 30 bytes at 62 absorb exactly
+ * that, which is why both halves of that span carry a width assert. */
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->hash) == 32,
                       "hash[] is 32 bytes on the wire");
-EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->reserved) == 30,
-                      "reserved[] is 30 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_len) == 2,
+                      "tlv_len is 2 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_hash) == 28,
+                      "tlv_hash is 28 bytes on the wire");
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->signature) == 64,
                       "signature[] is 64 bytes on the wire");
 
--- a/include/eos_secure_boot.h
+++ b/include/eos_secure_boot.h
@@ -106,8 +106,13 @@ int eos_secure_boot_verify_key(const uint8_t key_hash[32]);
 /**
  * @brief Lock debug interfaces (SWD/JTAG) permanently.
  * Only effective on real hardware with eFuse support.
+ *
+ * @return EOS_OK if the lock was written, otherwise the HAL error --
+ *         EOS_ERR_NOT_SUPPORTED on a board with no otp_write. Callers that
+ *         asked for the debug port to be closed must treat a non-OK result
+ *         as a failure to boot: the port is still open.
  */
-void eos_secure_boot_lock_debug(void);
+int eos_secure_boot_lock_debug(void);
 
 /**
  * @brief Update the anti-rollback counter in OTP.
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -124,7 +124,8 @@ if(VALGRIND)
                       test_multicore test_board_registry test_slot_manager
                       test_boot_log test_image_verify test_image_abi
                       test_recovery test_slot_size_bounds test_fw_transport
-                      test_tlv_auth)
+                      test_tlv_auth test_secure_boot_policy test_fdt_loader
+                      test_fw_decrypt)
         add_test(
             NAME valgrind_${TEST_NAME}
             COMMAND ${VALGRIND} ${VALGRIND_OPTS} $<TARGET_FILE:eboot_${TEST_NAME}>
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -90,6 +90,11 @@ add_executable(eboot_test_keystore unit/test_keystore.c)
 target_link_libraries(eboot_test_keystore PRIVATE eboot_core)
 add_test(NAME test_keystore COMMAND eboot_test_keystore)
 
+# --- test_secure_boot_policy: debug-lock policy enforcement ---
+add_executable(eboot_test_secure_boot_policy unit/test_secure_boot_policy.c)
+target_link_libraries(eboot_test_secure_boot_policy PRIVATE eboot_core)
+add_test(NAME test_secure_boot_policy COMMAND eboot_test_secure_boot_policy)
+
 # --- test_rollback: Anti-rollback security counter ---
 add_executable(eboot_test_rollback unit/test_rollback.c)
 target_link_libraries(eboot_test_rollback PRIVATE eboot_core)
--- a/tests/unit/test_ed25519.c
+++ b/tests/unit/test_ed25519.c
@@ -30,6 +30,7 @@ static int tests_passed = 0;
     static void name(void); \
     static void run_##name(void) { \
         printf("  %-50s ", #name); \
+        tests_run++; \
         name(); \
         tests_passed++; \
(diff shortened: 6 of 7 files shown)


srpatcha (commented):
## Review — eBoot#82 "fix(secure-boot): a debug lock that failed must not report a successful boot"
head: c5bb18192bfdd67174e69f86a961ef1fee69266a  author: Kartikey1306  ci: 24 checks pass

**Verdict:** The fix is right and the reasoning about `EOS_ERR_NOT_SUPPORTED` being the
common case, not the edge case, is the valuable part. The tests do not cover the thing the
PR changed, and the function being hardened is not reachable from the boot path.

### Findings
|---|---|---|---|---|
| 1 | Medium | `tests/unit/test_secure_boot_policy.c:88-121` | **The behavioural change is untested.** All three tests call `eos_secure_boot_lock_debug()` directly and assert its return value. None calls `eos_secure_boot()` with `cfg.lock_debug = true`, so the new Step-7 early return — the `attest_record(..., EOS_SBOOT_ERR_POLICY)` and `return EOS_SBOOT_ERR_POLICY` at `core/secure_boot.c:190-194` — never executes. The file's own docblock says "the debug-lock policy must be enforced, not merely attempted"; what it tests is that the helper reports, which is the "attempted" half. | Add a fourth test: build a `cfg` with `lock_debug = true` on a board with no `otp_write`, run a boot that would otherwise succeed, and assert `eos_secure_boot(...) == EOS_SBOOT_ERR_POLICY`. That is the regression that would catch a future revert. |
| 2 | Medium | `core/secure_boot.c:182` | `eos_secure_boot()` has **no production caller.** `git grep eos_secure_boot` across `master`, excluding its own file and `tests/`, returns one hit — a filename in `README.md:29`. The live path is `stage1/main.c:75` → `eboot_jump_to_app()` (`stage1/jump_app.c`), which parses, verifies integrity, verifies signature, checks rollback, applies the MPU and jumps — and never locks the debug port. So this PR closes a fail-open in a module nothing runs, while the path that does run has no debug-lock step at all. `docs/secure_boot_chain.md:143` documents `eboot_jump_to_app()` as step 12 of the chain, which confirms which one is intended to be live. | Merge this anyway — it is correct and cheap. Then open an issue: either wire `eos_secure_boot()` into Stage 1 or fold its policy steps into `eboot_jump_to_app()`. Two parallel orchestrators for the same responsibility is how one of them silently rots, which is what happened here. |
| 3 | Low | PR body, "Why this was never observable" | Stale. It says `core/secure_boot.c` is not in `CMakeLists.txt` and that this PR adds the line. #72 landed that (`2955938`), and `files.txt` confirms this PR no longer touches the top-level `CMakeLists.txt` at all. The `grep -c "core/secure_boot.c" CMakeLists.txt -> 0` evidence block no longer reproduces. | Update the body so the record is accurate; the finding it supports is still real. |
| 4 | Low | `tests/CMakeLists.txt:86` | Same insertion anchor as #80 — both add their triple immediately after `add_test(NAME test_keystore ...)`. Same author, both open, guaranteed conflict for whichever lands second. | Re-anchor one of them. |

### On the fix itself
The defect is exactly as described. `eos_hal_otp_write()` returns `EOS_ERR_NOT_SUPPORTED`
when the board supplies no `otp_write` hook, and the old `void eos_secure_boot_lock_debug(void)`
discarded it, so `lock_debug: true` was a silent no-op on every such board while
attestation recorded `EOS_SBOOT_OK`. Reusing `EOS_SBOOT_ERR_POLICY` (`include/eos_secure_boot.h:45`,
"Boot policy violation") rather than inventing a code is right. Recording the failed
attestation before returning matches how every other failure in the function behaves.

`void` → `int` on a public header is source- and ABI-compatible for existing callers, so
§23.2's compatibility contract is not engaged.

Worth noting for whoever wires finding 2: this fix means a board with `lock_debug: true`
and no `otp_write` will refuse to boot. That is the correct fail-closed behaviour and I am
not arguing against it — but today it is unreachable, and the day `eos_secure_boot()`
becomes live is the day it becomes an availability decision. `lock_debug` is set `true`
nowhere in the tree at present (`git grep lock_debug` finds only the struct member, the
two call sites, and `test_secure_boot.c:140` setting it `false`), so there is time to
decide whether the right response is a hard stop or a route to recovery per §8.1.

### Architecture conformance
Conforms. §21 Tier 1 (`eBoot`, Foundation). §5.1 — "eBoot keeps the trusted computing base
minimal and auditable": a policy step that reports success without having taken effect is
the opposite of auditable, and this removes one. §8.1 requires "Explicit separation between
implemented, experimental and planned security features", and §28 grades `Implemented` as
requiring "Code and functional tests" — finding 1 is that gap, finding 2 is a sharper
version of the same one. No dependency edges added.

### Proposed changes
```
core/secure_boot.c            keep as-is — the fix is correct
tests/unit/test_secure_boot_policy.c
                              + test_secure_boot_returns_policy_error_when_lock_fails()
                                driving eos_secure_boot() end to end
tests/CMakeLists.txt          re-anchor away from #80's insertion point
PR body                       drop the CMakeLists claim (#72 landed it)
follow-up issue               reconcile eos_secure_boot() with eboot_jump_to_app()
```

### Not checked
- Nothing was built or run. `build clean`, `ctest 20/20`, `pytest 30 passed` are the
  author's figures; I did not reproduce them, and I did not confirm that the new test
  fails against unmodified `secure_boot.c` as claimed.
- `checks.txt` shows 24 green checks on this head — CI's signal, not mine.
- I did not review `tests/unit/test_secure_boot.c` (added by #72) for overlap with the new
  file beyond the naming collision the author already avoided.
- Whether any out-of-tree board sets `lock_debug = true` is unknown; finding 2's
  availability note covers in-tree configuration only.


---
*Automated architecture review of `c5bb18192bfd` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eBoot#82 "fix(secure-boot): a debug lock that failed must not report a successful boot"
head: eafe67c5d3c59e1c39355df5f2d04ec5845525fe  author: Kartikey1306  ci: 23 pass, 1 skipped (Create GitHub Release)

**Verdict:** The fix is right and the follow-up commit closes the coverage gap properly —
including the counter-check that stops the new test passing for the wrong reason. One new
High finding, in the same function and of the same class the PR exists to fix.

### What I ran

Built `refs/pull/82/head` from a clean export, host gcc,
`-DEBLDR_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug`:

```
cmake exit 0   build exit 0
ctest --no-tests=error --output-on-failure --timeout 120  ->  exit 0, 22/22 passed
   test_secure_boot_policy  Passed   (6/6 within it)
```

`origin/master` (22d8f8b) under the same procedure fails to build and runs 0 of 21 tests —
`include/eos_image.h:135` references a removed `reserved` member and
`core/ed25519_verify.c:338` redefines `point_is_identity`. This branch is stacked on **#94**,
which repairs both; that is why `gh pr diff` shows `ed25519_verify.c`, `eos_image.h` and
`test_ed25519.c` here. This PR's own delta is `0886ac0` + `eafe67c`, over
`core/secure_boot.c`, `include/eos_secure_boot.h`, `tests/CMakeLists.txt` and
`tests/unit/test_secure_boot_policy.c`.

### Previous findings (head `c5bb1819`)

| Was | Now |
|---|---|
| M1 the behavioural change was untested — every test drove the helper, none drove `eos_secure_boot()` | **Fixed, and better than asked.** `test_secure_boot_refuses_when_the_debug_lock_cannot_be_taken` drives the whole function to `EOS_SBOOT_ERR_POLICY`, and `test_the_same_image_boots_when_no_debug_lock_is_asked_for` is the counter-check that proves steps 1–6 actually reach step 7 rather than the first test passing off an unrelated failure. `test_secure_boot_proceeds_when_the_debug_lock_succeeds` closes the third corner. The `stage_bootable_image()` comment about `EOS_IMG_FLAG_HASH_SHA256` being required or step 2 takes the CRC32 branch is the sort of thing that is usually rediscovered rather than written down. |
| M2 `eos_secure_boot()` has no production caller | **Still open.** `git grep 'eos_secure_boot('` on this head returns the definition and the declaration, nothing else. Restated below, not re-argued. |
| L3 PR body claims `core/secure_boot.c` is not in `CMakeLists.txt` | **Still stale.** `origin/master:CMakeLists.txt:110` already lists it, and this PR does not touch that file. |
| L4 `tests/CMakeLists.txt` anchor clash with #80 | **Resolved** — #80 moved its block to after `test_ecc`; this one keeps the `test_keystore` anchor. No conflict. |

### Findings
|---|---|---|---|---|
| 1 | High | `core/secure_boot.c:104-118` | **Step 4 is the same defect this PR is fixing, two steps earlier, and is untouched.** "Verify signing key against OTP root-of-trust" reads the OTP hash, and then: if `eos_hal_otp_read()` returns anything but `EOS_OK` the failure is discarded and boot continues; if it succeeds and the hash is provisioned, the body of the `if` is `/* In a full implementation, extract key hash from TLV and compare */` — no comparison happens. So a device whose root of trust *is* provisioned boots an image signed by any key the image itself carries, and `attest_record(..., EOS_SBOOT_OK)` records a clean secure boot. `.ai/security.md`: "A verification step that cannot run must fail, not pass"; "Attestation must not record OK for a boot in which a policy step was skipped." This is pre-existing, not introduced here — but it is in the function this PR edits, and the PR's own thesis is the argument for fixing it. | Not this PR's scope to implement TLV key-hash extraction. Do two things that are: fail on a non-`EOS_OK` `otp_read`, and make the unimplemented branch `return EOS_SBOOT_ERR_SIGNATURE` when OTP is provisioned, so a provisioned device fails closed instead of booting on an unverified key. If that is judged too large, the minimum is that step 4 stops being labelled a verification — §8.1 requires "explicit separation between implemented, experimental and planned security features", and a comment inside an `if` is not that separation. |
| 2 | Medium | `core/secure_boot.c:71` | Carried forward from `c5bb1819`, still true at this head: `eos_secure_boot()` has no production caller. The live path is `stage1/main.c:75` → `eboot_jump_to_app()`, which never locks the debug port. Two orchestrators for one responsibility is how one of them rots — which is what produced both this fail-open and finding 1. | Merge this anyway; then open an issue to either wire `eos_secure_boot()` into Stage 1 or fold its policy steps into `eboot_jump_to_app()`. Note that finding 1 is what makes this urgent rather than tidy: the unreachable orchestrator is the one that would be reached first if anyone wired it up as-is. |
| 3 | Low | `tests/unit/test_secure_boot_policy.c:30-37, 232` | `TEST()` does not increment `tests_run` and `main()` hardcodes `tests_run = 6`, so a test added to this file but not wired into `main()` is skipped silently with a zero exit. #94 — the base commit of this branch — removed exactly this pattern from `tests/unit/test_ed25519.c`, where a hardcoded `tests_run = 11` was masking two uncalled tests. #80 has the same finding against its new file. | `tests_run++;` inside the `TEST()` macro, drop the assignment in `main()`. Matches `tests/unit/test_ed25519.c:29-36` on this branch. |
| 4 | Low | `tests/CMakeLists.txt:93-97` | `test_secure_boot_policy` is not in the Valgrind `foreach(TEST_NAME ...)` list at `tests/CMakeLists.txt:121-131`, so it gets no memory-safety run. The list is hand-maintained and now misses 6 of 22 registered tests. | Add it. Separately, derive the list from the registered tests — nothing fails today when a name is forgotten. |
| 5 | Low | PR body | Two claims no longer reproduce: `grep -c "core/secure_boot.c" CMakeLists.txt -> 0` (it is 1 on master, landed by #72) and "Added the one line that builds it". The verification table also predates the follow-up commit — it says `ctest 20/20`, this head is 22/22. | Refresh the body. The finding it supports is real; the evidence quoted for it is not. |

### On the fix itself

Correct, and I confirmed the mechanism rather than taking the body's word. `eos_hal_otp_write()`
returns `EOS_ERR_NOT_SUPPORTED` when the board supplies no `otp_write`, the old
`void eos_secure_boot_lock_debug(void)` discarded it, so `lock_debug: true` was a silent no-op
on every such board while attestation recorded `EOS_SBOOT_OK`. Reusing the existing
`EOS_SBOOT_ERR_POLICY` rather than inventing a code is right, and recording the failed
attestation before returning matches every other failure path in the function.

`void` → `int` on the public prototype is source- and ABI-compatible for existing callers, so
§23.2's compatibility contract is not engaged. `lock_debug` is still set `true` nowhere in the
tree (`git grep lock_debug` on this head: the struct member, the two call sites, and
`test_secure_boot.c:140` setting it `false`), so the availability consequence of failing closed
is not yet live on any in-tree board.

### Architecture conformance

Conforms. §21 Tier 1 (`eBoot`, Foundation) — right repo. §5.1 "eBoot keeps the trusted
computing base minimal and auditable": a policy step reporting success without having taken
effect is the opposite of auditable, and this removes one of two — finding 1 is the other.
§28 grades `Implemented` as requiring "code and functional tests"; the debug-lock policy now
meets that bar and step 4 does not. No dependency edges added.

The master design has no gap here that I can see: §8.1 and §14.1 already require what
findings 1 and 2 ask for. Nothing appended to `proposals/`.

### Proposed changes

```
core/secure_boot.c:104-118   fail on non-EOS_OK otp_read; fail closed when OTP is
                             provisioned and the key-hash comparison is not implemented
tests/unit/test_secure_boot_policy.c:30-37,232
                             tests_run++ in TEST(); drop tests_run = 6
tests/CMakeLists.txt:126     add test_secure_boot_policy to the valgrind foreach
PR body                      drop the CMakeLists claim, refresh 20/20 -> 22/22
follow-up issue              reconcile eos_secure_boot() with eboot_jump_to_app()
```

Only finding 1 is worth blocking on, and only if you judge it in scope; the other four are
one-liners.

### Not checked

- Host build and `ctest` only — gcc, x86-64, Debug. No cross builds, no sanitizer build, no
  valgrind, no `pytest tests/`. The 23 green checks in `checks.txt` cover those; that is CI's
  evidence, not mine.
- I did not verify the claim that the new tests fail to compile against unmodified
  `secure_boot.c` ("invalid operands to binary expression ('void' and 'int')"). It is
  plausible from the signature change, but I did not run it.
- I did not review `tests/unit/test_secure_boot.c` (from #72) for behavioural overlap with the
  new file, only for the naming collision the author already avoided.
- `eos_secure_boot_lock_debug()` does not read the fuse back after writing it; whether a
  successful `otp_write` implies a blown fuse on real silicon is a HAL contract question I
  cannot settle from the tree.
- Step 4's `secure_compare()` usage and the attestation chain hash were read, not tested.


---
*Automated architecture review of `eafe67c5d3c5` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eBoot#82 "fix(secure-boot): a debug lock that failed must not report a successful boot"
head: 81bfe0b7641067f59d1d8c06c2e15785b3e6f49c  author: Kartikey1306  ci: pass

**Verdict:** The core change is right and well tested — a discarded `eos_hal_otp_write()` result meant `lock_debug: true` booted with SWD/JTAG open and recorded `EOS_SBOOT_OK`, and step 7 now fails closed. But the PR also carries an unannounced step-4 change that makes every OTP-provisioned device refuse to boot, and a two-name addition to the Valgrind test list that breaks CMake generation wherever valgrind is installed.

### Findings

|---|---|---|---|---|
| 1 | High | `tests/CMakeLists.txt:127-128` | `test_fdt_loader` and `test_fw_decrypt` are added to the Valgrind `foreach`, but no `eboot_test_fdt_loader` / `eboot_test_fw_decrypt` target exists in this repo. `$<TARGET_FILE:...>` then fails at generate time, so `cmake -B build` aborts on any host that has valgrind. Reproduced against this head with `-DVALGRIND=/usr/bin/true`: `CMake Error at tests/CMakeLists.txt:129 (add_test): No target "eboot_test_fdt_loader"` … `CMake Generate step failed.` This is invisible in the PR checks because no required job installs valgrind — but `.github/workflows/weekly.yml:59-73` ("Valgrind Memcheck") installs valgrind and then runs `cmake -B build`, so that job breaks on merge. | Drop the two names. `test_secure_boot_policy` is the only new target this PR adds and it is correct to list. |
| 2 | High | `core/secure_boot.c:123-140` | Step 4 now returns `EOS_SBOOT_ERR_SIGNATURE` unconditionally whenever `require_signature` is set and the OTP anchor is non-zero. A device whose root of trust *is* provisioned — the intended production configuration — can no longer boot any image at all. The same holds for `rc != EOS_OK`, which includes `EOS_ERR_NOT_SUPPORTED` on a board with no `otp_read` hook. Failing closed is the right direction (`.ai/security.md`, "Fail closed"), but this is a device-bricking behaviour change that is not in the title, not in the PR body's "The change" section, appears only inside a code comment, and by the author's own note has no test. | Split it into its own PR. It needs an explicit decision, a release note, and either the TLV comparison it is standing in for or a config gate. Landing it inside a debug-lock fix hides it. |
| 3 | Medium | `core/secure_boot.c:126,136` | Step 4's two refusals record and return `EOS_SBOOT_ERR_SIGNATURE`, which is exactly what step 3 returns for a genuinely bad signature. The condition is therefore unobservable from the result code and, as the note at `tests/unit/test_secure_boot_policy.c:669-687` says, untestable — the obvious test passed against the unfixed code too. `EOS_SBOOT_ERR_KEY_MISMATCH` (`include/eos_secure_boot.h:42`, "Key hash doesn't match OTP") already exists and is the code for this. | Return `EOS_SBOOT_ERR_KEY_MISMATCH` from both step-4 refusals. The test that could not be written then becomes writable with the existing unsigned fixture. |
| 4 | Medium | PR body | Two claims are not supported. (a) "`core/secure_boot.c` is not in `CMakeLists.txt` … `grep -c` → 0 … Added the one line that builds it" — #72 merged as `2955938`, `core/secure_boot.c` is at `CMakeLists.txt:110` on `origin/master`, and this PR does not touch the root `CMakeLists.txt` at all (`files.txt` confirms). (b) "`ctest --no-tests=error` **20/20 pass**" — this head actually runs 22 tests. Per the brief, an assertion carried without matching output is itself the finding. | Re-run against current `origin/master` and restate. The stale #72 paragraph should be deleted, not corrected in place. |
| 5 | Low | `core/ed25519_verify.c`, `include/eos_image.h`, `tests/unit/test_ed25519.c` | Three changes unrelated to the debug lock ride along. They are not wrong — see "Not checked"/evidence below, `origin/master` does not compile without them — but nothing in the PR says the PR is also the master-build fix, so a reviewer looking only at the title will not know that reverting it re-reds the tree. | One line in the body naming the two `origin/master` compile errors this also fixes. |

### Architecture conformance

Conforms. Master design §8.1 requires "explicit separation between implemented, experimental and planned security features"; the step-4 comment does exactly that, and §5.1's "eBoot keeps the trusted computing base minimal and auditable" is served by making a policy step's failure visible rather than discarded. No dependency direction is touched: `core/secure_boot.c` calls only `eos_hal_*` and `core/` peers, all at or below its own tier. Tier placement is correct — eBoot is Tier 1 Foundation (§21) and secure-boot policy is eBoot's own responsibility per §8.

The `.ai/architect.md` rule "restructure and change behaviour in the same commit" is stretched by finding 2, not by findings 5: the ed25519 and `eos_image.h` edits are build repair, the step-4 edit is new behaviour.

### Proposed changes

1. `tests/CMakeLists.txt` — revert the Valgrind list to add one name only:
   ```diff
   -                      test_tlv_auth test_secure_boot_policy test_fdt_loader
   -                      test_fw_decrypt)
   +                      test_tlv_auth test_secure_boot_policy)
   ```
   Verify with `cmake -B build -DEBLDR_BUILD_TESTS=ON -DVALGRIND=/usr/bin/true`, which currently fails and must configure clean.
2. `core/secure_boot.c` — change both step-4 `EOS_SBOOT_ERR_SIGNATURE` returns to `EOS_SBOOT_ERR_KEY_MISMATCH`, then add the test that was previously impossible: provisioned anchor + `require_signature = true` + valid unsigned fixture → expect `EOS_SBOOT_ERR_KEY_MISMATCH`, which fails against the unfixed code because step 3 returns `-5`.
3. Move the step-4 hunk to its own PR, or keep it and say in the body that it stops provisioned devices booting until TLV extraction lands.
4. Correct the two PR-body claims.

Findings 1 and 3 are the only ones that need to land before merge; 1 because it breaks a scheduled job, 3 because it is two token changes and it unblocks the test the author wanted to write.

### Not checked

- **Not verified:** nothing was run on hardware. `eos_hal_otp_write()` failure and `EOS_ERR_NOT_SUPPORTED` were exercised only through the simulated board in `test_secure_boot_policy.c`; whether a real eFuse write reports failure the way the test assumes is untested here.
- **Not verified:** the Cortex-M `#if defined(__ARM_ARCH)` block in `eos_secure_boot_lock_debug()` is still entirely commented out, so on ARM the function reports `EOS_OK` after writing the OTP fuse without having touched DHCSR/DBGMCU. Whether the OTP fuse alone actually closes SWD/JTAG on the supported boards is a hardware question I cannot answer from here, and no test covers it.
- **Not verified:** the eight low-order encodings in `k_low_order[]` and the three in `k_non_canonical[]` were not independently recomputed; the suite passing only shows they are all rejected, not that they are the points the comments name.
- **Not verified:** the claim "Measured against 13a7a02 … 16 of the 64 (key, R) pairs were accepted" — I did not reproduce it.
- **Not checked:** interaction with #72 and #88 beyond confirming #72 is already merged and does not collide with this diff.

### Evidence

Run locally against `origin/master` (`22d8f8b`) and this head, from read-only `git archive` extractions — no repository was modified:

```
origin/master :  cmake -B build -DCMAKE_BUILD_TYPE=Debug -DEBLDR_BUILD_TESTS=ON   -> OK
                 cmake --build build                                              -> FAIL, 12 errors
                 gcc -c -Iinclude core/ed25519_verify.c                            -> FAIL
       (previous definition at core/ed25519_verify.c:281)

81bfe0b7      :  cmake -B build ...                                               -> OK
                 cmake --build build -j4                                          -> OK
                 ctest --no-tests=error                                           -> 22/22 passed
                 cmake -B build-vg ... -DVALGRIND=/usr/bin/true                   -> FAIL (finding 1)
```

So `origin/master` is red on two counts today and this branch is the thing that makes it green again — worth saying plainly, because finding 1 is the only reason it cannot land as it stands.

Separately: `tests/unit/test_ed25519.c` on `origin/master` hard-codes `tests_run = 11` while `main()` makes 13 `run_*` calls (one of them duplicated), and `test_ed25519_low_order_R_with_a_valid_key_is_not_a_forgery` is defined but never called. This PR fixes both by counting in the `TEST` macro. That is a real repair, not bookkeeping — a suite that reports its own total cannot detect a test it stopped running.


---
*Automated architecture review of `81bfe0b76410` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
## Correcting the body — two claims no longer reproduce

Finding 5 is right, and both are my own stale evidence rather than a change in the code.

**"`grep -c "core/secure_boot.c" CMakeLists.txt` → 0"** — it is `1` on `master`. #72 landed that line (`2955938`) after I wrote the body, and this PR no longer touches the top-level `CMakeLists.txt` at all. The finding that block supported — that `core/secure_boot.c` was uncompiled and therefore untestable — was true when written and is not the reason to take this PR now.

**"Added the one line that builds it"** — likewise no longer this PRs doing.

**"`ctest 20/20`"** — this head is **22/22**, and `test_secure_boot_policy` is 7/7 within it. That number predates the follow-up commits.

The current state, run just now on this head:

```
ctest --no-tests=error --timeout 120   ->  22/22 PASS
eboot_test_secure_boot_policy          ->  7/7 PASS
```

## On finding 1 — step 4

Fixed in the latest commit, and it is the more important half of this PR now. A non-`EOS_OK` `otp_read` fails `EOS_SBOOT_ERR_SIGNATURE`, and a **provisioned** anchor that nothing compares against refuses the boot rather than proceeding. An unprovisioned (all-zero) anchor is deliberately unchanged — there is nothing to check against and refusing would brick every board that has not been provisioned.

**There is no test for those two refusals, and the file says so rather than leaving it to be found.** Reaching step 4 requires passing step 3, which is a real Ed25519 signature checked against the keystore. I wrote the obvious test first and it was worthless: with `require_signature = true` and an unsigned fixture the boot fails at **step 3** and returns the same `EOS_SBOOT_ERR_SIGNATURE` step 4 returns, so it passed against the unfixed code too. I confirmed that by reverting step 4 and watching it still pass, then deleted it. A test that cannot fail is worse than none.

The fixture is buildable — the keystore ships RFC 8032 TEST 1s public key and the matching private key is in the RFC — but that machinery is #88s (`tools/gen_signed_image_fixture.py`). Worth doing once #88 lands; I would rather say that than ship a green test that proves nothing.

Findings 3 and 4 are done here (`tests_run` derived; `test_secure_boot_policy`, `test_fdt_loader` and `test_fw_decrypt` added to the Valgrind list). **#95** fixes both as classes across the whole suite — 18 of 21 suites hardcoded their total, and the Valgrind list had drifted to 17 of 21 — with a guard so neither returns.

PR #89, opened 2026-09-01
test(ed25519): hold both repos' verifiers to one shared contract

Kartikey1306:
> **Stacked on #86.** Review the last commit. Without #86's fix this test fails, which is rather the point.

## The gap this closes

eos and eBoot each carry their own Ed25519 verifier. Different code, **opposite return conventions** — `ed25519_verify()` returns `1` to accept, `eos_ed25519_verify()` returns `EOS_OK` — doing the same job on the same wire format.

For a while only eos rejected low-order public keys. Nothing in either repo could notice: there is no shared build, and the two implementations are far too different for a source diff to say anything useful. That divergence is *why* #73 survived in eBoot after eos had already fixed the identical defect in `services/crypto/src/ed25519_verify.c`.

#86 closes the hole. It does not stop the next crypto fix in either repo from failing to reach the other.

## What is shared

The implementations can't be. The **data** can. `tests/vectors/ed25519_contract_vectors.h` is 76 vectors of pure test data plus a SHA-256 over them:

| count | vectors |
|---|---|
| 64 | low-order keys — the eight points of order dividing 8, each over eight messages |
| 3 | RFC 8032 §7.1 signatures that **must verify** |
| 6 | those three with one bit flipped in R or in S |
| 3 | those three with `S + L`, non-canonical per RFC 8032 §5.1.7 |

Eight messages per key rather than one because a key of order *n* makes `(R = identity, S = 0)` verify whenever *n* divides `SHA-512(R‖A‖M)` — about one message in *n*. **A regression test pinned to a single message passes against unfixed code** for the order-4 and order-8 points.

The intended twin is `tests/test_ed25519_contract.c` in eos, compiling the byte-identical header. Each side runs its own verifier and prints the digest, so a change to one copy that does not reach the other surfaces as two different digests — the thing a diff of two unrelated implementations could never give you.

## The corpus is generated, not transcribed

```
python3 tools/gen_ed25519_contract_vectors.py > tests/vectors/ed25519_contract_vectors.h
```

so it can be regenerated and re-checked rather than trusted. The generator carries the reasoning for each group of vectors.

## The positive vectors are load-bearing

A verifier that refuses everything satisfies all 73 negative vectors. The three RFC 8032 signatures are the only thing standing between "hardened" and "broken", so the test fails explicitly if the corpus ever loses its accept cases, not just if a vector misbehaves.

## Validation

```
on this branch (with #86)        ->  76/76,  3 accepted, 73 refused
against unfixed origin/master    ->  28 of 76 wrong, exit 1
ctest                            ->  22/22 passed
digest  1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2
```

The 28 failures on master are the low-order forgeries that land; the other 36 low-order vectors are ones where `k` happened not to be divisible by the key's order, which is exactly the sampling behaviour the eight-messages-per-key design is there to defeat.

## Follow-up

The eos side is a separate PR against that repo — same header, same digest, its own driver for the `1`/`0` convention. This one is useful on its own (it is a real test of eBoot's verifier); it becomes a drift guard once both are in.


Kartikey1306:
The eos twin is now open: **embeddedos-org/eos#122**, same corpus, same digest `1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2`.

It earned its keep immediately. Running this corpus against eos's verifier failed on three vectors that eBoot passes: eos had only ref10's partial `signature[63] & 224` check, so every `S` in `[L, 2^253)` was accepted and `(R, S)` and `(R, S + L)` both verified — signature malleability. eBoot's `s_is_canonical()` has covered that all along. Fixed on the eos side in #122.

That is the case for the corpus in one sentence: neither implementation was badly designed, one just had a check the other lacked, and nothing in either repo could tell you.

srpatcha (commented):
## Review — eBoot#89 "test(ed25519): hold both repos' verifiers to one shared contract"
head: a59676a8797a58451ae6a26839979676aa313316  author: Kartikey1306  ci: pass

**Verdict:** The idea is right and it has already paid for itself — running this corpus against eos found a real malleability hole (`S` in `[L, 2^253)` accepted) that no source diff of two unrelated implementations would ever have surfaced. As a test of eBoot's verifier the corpus is good work. But the drift guard the PR is named for is not implemented: the digest is printed and never asserted, and it is computed by the generator over the generator's own output, so a generator that drifts in one repo produces a self-consistent header, a different number, and a green test. Nothing fails and nothing compares.

### Findings
|---|---|---|---|---|
| 1 | High | `tests/unit/test_ed25519_contract.c:36` | The digest is `printf`'d and never checked. `main()` prints `EOS_ED25519_CONTRACT_DIGEST`, loops the vectors, and returns 1 only on a per-vector mismatch or a zero-positives corpus. There is no comparison against any expected value, and `EOS_ED25519_CONTRACT_DIGEST` comes from the build-tree header that `tools/gen_ed25519_contract_vectors.py` just wrote — a hash the generator took over its own output. So if eos's copy of the generator drifts, eos regenerates a self-consistent corpus with a new digest and **its test still passes**; eBoot's also passes; both suites stay green and the divergence is visible only to a human reading two CI logs in two repos and noticing two hex strings differ. The claim is asserted in three places and implemented in none: the PR body ("surfaces as two different digests"), `test_ed25519_contract.c:26-28` ("asserts the same digest … Both files should always report the same digest as the one below" — there is no digest literal below, or anywhere in the file), and the generator's own emitted header comment at `tools/gen_ed25519_contract_vectors.py:115` ("Both repos compile this identical file and assert the digest below"), plus its docstring at `:19-21` ("the build fails and the digests visibly stop matching" — the build does not fail). The unused `#include <string.h>` at `test_ed25519_contract.c:29` — no string function is called — looks like the assertion was intended and lost. | Pin the digest as a literal in committed source, in both repos: `#define EOS_ED25519_CONTRACT_EXPECTED "1059febe…282a2"` in the driver, then `if (strcmp(EOS_ED25519_CONTRACT_DIGEST, EOS_ED25519_CONTRACT_EXPECTED) != 0) { printf("[FAIL] corpus digest changed: generator drift or an unsynchronised edit\n"); return 1; }`. That turns a generator change into a failing build in whichever repo it lands in, which forces the deliberate two-repo update this PR exists to force. It also closes two holes the current guard misses for free: a corpus that shrinks from 76 vectors to 40, and one that drops from 3 accept vectors to 1 — both invisible today, because `EOS_ED25519_CONTRACT_COUNT` is generated too and the `positives == 0` check at `:70-76` only fires when the last positive goes. |
| 2 | Medium | `tests/CMakeLists.txt:84-105` and the absent `tests/vectors/ed25519_contract_vectors.h` | The corpus is generated into `${CMAKE_CURRENT_BINARY_DIR}/vectors/` and not committed, so what the TCB's signature verifier was tested against is not in the repository and not in any release artifact. From a tag you cannot tell which 76 vectors ran. `.ai/security.md` requires that security claims rest on a check that was actually run and that the remaining risks be written down; a build-time-generated corpus leaves no record of either. The CMake comment gives the reason — "the identical file would otherwise be checked in to both eos and this repo and kept in step by hand" — and that concern is real but is exactly what finding 1's pinned digest removes: with the digest asserted, the two copies cannot drift apart in silence, so committing them is safe. Related: `find_package(Python3 COMPONENTS Interpreter REQUIRED)` makes Python 3 a hard configure-time dependency of the whole eBoot test suite. `cmake` now fails outright on any host or container image without it, including minimal cross-build images, for a target that is otherwise pure C. | Commit `tests/vectors/ed25519_contract_vectors.h` in both repos, keep the generator as the way it is *reproduced* and reviewed, and let finding 1's pinned digest be what keeps the two copies honest. That drops the `find_package(Python3 … REQUIRED)` and the `add_custom_command`, and makes the corpus auditable from a tag. If generating stays preferred, at minimum make Python optional — `find_package(Python3 COMPONENTS Interpreter)` and skip registering `test_ed25519_contract` when it is missing — so a host without Python loses one test rather than the ability to configure. |
| 3 | Medium | PR body, "**Stacked on #86.** Review the last commit"; `files.txt` | The stack is understated by two PRs and one subsystem. The diff carries `core/fdt_loader.c` (+100/-15), `tests/unit/test_fdt_loader.c` (+375), the `tests/CMakeLists.txt` FDT hunk and the `CMakeLists.txt` line adding `core/fdt_loader.c` to `eboot_core` — that is all of #84 and #85, device-tree parsing, unrelated to Ed25519 — plus #86's `core/ed25519_verify.c` (+52) and `tests/unit/test_ed25519.c` (+85/-1). So a reviewer told to look at the last commit will not know that three unmerged PRs and a second subsystem are riding along, and if #84 or #85 change during review this branch silently carries stale copies of them. Compounding it, the org remote carries a third equivalent implementation of #86's guard on the branch `fix/ed25519-low-order-keys` at `8b88125` (`public_key_is_valid_subgroup`, srpatcha, unmerged — see the #86 review), so whichever of the three lands will conflict with the copy this branch carries. `.ai/architect.md` asks that restructuring and behaviour change not share a commit; the same reasoning applies to two unrelated subsystems sharing a branch. | Declare the full stack in the body — "stacked on #84 → #85 → #86" — and separate the subsystems: drop the FDT files, which belong to #84/#85 and have nothing to do with Ed25519. The Ed25519 dependency is real and has to stay until one of the three guards merges, since `master` (`13a7a02`) still has no subgroup check and this corpus genuinely fails against it. Once one lands, rebase onto `master` and reduce this PR to `tools/gen_ed25519_contract_vectors.py`, `tests/unit/test_ed25519_contract.c` and its `tests/CMakeLists.txt` hunk. |
| 4 | Low | `tests/CMakeLists.txt:131-137`, the Valgrind `foreach` | `test_ed25519_contract` is added to the Valgrind list but `test_fdt_loader` — added 6 lines above in the same diff — is not, even though the FDT loader is the newly-compiled untrusted-input parser in this change set and the one most likely to have a memory error. Inconsistent within a single PR. | Add `test_fdt_loader` to the same `foreach`, or state why it is excluded. |

### Architecture conformance
Conforms. §21 Tier 1 — Foundation. `tests/unit/`, `tests/vectors/` and `tools/` are correct homes; no `core/`, `stage1/`, `hal/` or `boards/` change belongs to this PR's own commit, and §5.1's minimal-TCB requirement is untouched since nothing here links into the boot image. No include, link line or `target_link_libraries` entry points up a tier — the new test links `eboot_core`, which is downward. §14.1 "Use reviewed cryptographic libraries; do not invent cryptographic primitives" is served rather than strained: this adds no primitive, only vectors. §8.1's "Explicit separation between implemented, experimental and planned security features" is what finding 1 is ultimately about — the drift guard currently reads as implemented and is planned.

The structural problem is the one this PR is a response to, and it is a gap in the master design. §14.1 requires reviewed libraries and integrated key management across eBoot, eSec, eOTA and release signing, but nothing says a cryptographic primitive has one implementation in the ecosystem, and nothing names an owner. Meanwhile §21 puts eBoot in Tier 1 and eSec in Tier 2, so the obvious deduplication — eBoot consuming eSec's verifier — points *up* a tier and is forbidden by §5.1. The design therefore rules out the clean fix while being silent on the alternative, which is how the same Ed25519 low-order bypass came to exist in two repos and be fixed in only one for weeks. This PR's shared-corpus approach is a sound answer to that constraint and deserves to be the documented one. Proposal appended to `.ai/autoreview/proposals/2026-09.md`.

### Proposed changes
1. Assert the digest against a committed literal (finding 1). This is the change that makes the PR do what its title says; everything else is secondary. It also subsumes the corpus-count and positive-count gaps.
2. Commit the corpus header and drop the hard Python dependency (finding 2) — or make Python optional if generating stays.
3. Drop the FDT files and declare the full stack (finding 3). Reducing to this PR's own three files has to wait until one of the three competing Ed25519 guards merges — `master` still has none, so the corpus legitimately needs #86 or an equivalent underneath it.
4. Add `test_fdt_loader` to the Valgrind list (finding 4).
5. Correct the three places that describe the assertion as existing: the PR body, `test_ed25519_contract.c:26-28`, and `tools/gen_ed25519_contract_vectors.py:19-21,115`. Once finding 1 lands they become true rather than needing to be softened.

Worth stating plainly for the maintainers: the malleability defect this corpus found in eos (`S` in `[L, 2^253)` accepted, `(R, S)` and `(R, S + L)` both verifying, reported in the 2026-09-01 comment and fixed in eos#122) is the strongest argument in this PR, and it is independent of the drift-guard weakness. The corpus works. It should land — with the digest actually asserted.

### Not checked
- **Nothing was executed.** No `ctest`, no generator run, no corpus. Reproducing requires checking out the PR head, which the run brief forbids. Every number is the author's: `76/76`, `3 accepted, 73 refused`, `28 of 76 wrong` on unfixed master, `ctest 22/22`, and the digest `1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2`. I did not verify that digest, did not verify the corpus is 76 vectors, and did not verify the 64/3/6/3 breakdown.
- **The generator's determinism claim I checked only by reading its imports.** `tools/gen_ed25519_contract_vectors.py` imports `hashlib` and `sys` and nothing else — no `random`, no `time`, no environment or filesystem input — so the claim that both repos derive the same digest from the same generator holds on inspection. I did not run it twice to confirm, and I did not check for dict- or set-ordering dependence in `vectors()`.
- **The eos side was not examined at all.** eos#122 is asserted to carry the byte-identical header, the same digest and a driver for the `1`/`0` convention. I did not open the eos working tree, did not confirm the twin exists, and did not confirm the malleability fix. Finding 1's severity rests on the mechanism, not on the state of eos — a pinned digest is needed whatever eos#122 contains.
- **The low-order vector set was not audited here.** My #86 review found that repo's `k_low_order[8]` omits three of the eight canonical low-order encodings and includes one non-member. Whether this PR's generated 64-vector group has the same gap I could not tell, because the corpus is not in the diff — only the generator that produces it, and I did not trace its point construction. **This is the one finding I would most expect to also apply here and cannot rule out.**
- **21 of 22 checks pass**; `Create GitHub Release` reports `skipping`, expected on a PR. No required check failing. Note the green result was produced with the `add_custom_command` in place, so CI's runner has Python 3 — which is why finding 2's configure-time dependency is invisible in `checks.txt`.
- **`mergeStateStatus: BLOCKED`, `mergeable: MERGEABLE`.** No merge or rebase attempted. `tests/CMakeLists.txt` and `CMakeLists.txt` collide with #84, #85, #88 and #90; expect conflicts in all four.


---
*Automated architecture review of `a59676a8797a` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eBoot#89 "test(ed25519): hold both repos' verifiers to one shared contract"
head: ea9bc23809df3f35f5ba4c09ad1429ecd91c20c0  author: Kartikey1306  ci: pass

**Verdict:** The three findings from the review of `a59676a8` are properly addressed — the corpus is committed, the Python configure-time dependency is gone from the test suite, and the branch is rebased down to its own three files. The corpus itself is sound: I recomputed the digest, the orders and the forgery distribution independently, and every number checks out. But the drift guard still does not do what the file says it does, and I can demonstrate that. My previous recommendation is what caused this — it named `strcmp` against a literal, and that is not sufficient.

### Findings

|---|---|---|---|---|
| 1 | High | `tests/unit/test_ed25519_contract.c:53-73` | The pinned digest does not cover the corpus it ships with. `EOS_ED25519_CONTRACT_DIGEST` is a `#define` string literal *inside* `tests/vectors/ed25519_contract_vectors.h`, written there by the generator. The test compares one literal against another literal. Nothing hashes the vector bytes, so an edit to the committed header changes the data and leaves the digest it is compared against untouched. Demonstrated below: I changed one byte of the order-1 identity public key — the vector that pins the Critical secure-boot bypass — and the suite reported `[PASS] all 76 contract vectors behaved as specified`. The guard fires only when someone re-runs the generator; the file it protects is the one it cannot see. `tests/CMakeLists.txt` now states the digest is what makes committing the corpus safe ("neither copy can change without failing"), and that is the load-bearing claim for the whole design. Both repos share the limitation — eos#122's driver at `tests/test_ed25519_contract.c:47-70` is the same three `#define`s and the same `strcmp`. | Recompute at runtime. `eos_sha256_init/update/final` is already in `eos_crypto_boot.h`, which this file already includes. The generator's serialisation is exact and simple (`tools/gen_ed25519_contract_vectors.py:95`): for each vector in order, `public_key[32] ‖ signature[64] ‖ message[message_len] ‖ (uint8_t)expect_accept`. Hash that over `eos_ed25519_contract_vectors[]`, hex-encode, and compare to `EOS_ED25519_CONTRACT_EXPECTED`. That makes the pin cover the data instead of a sibling `#define`, and it keeps every property the current version has. Same change on the eos side. |
| 2 | Low | `tools/gen_ed25519_contract_vectors.py:13-15`, the emitted header comment at `:109`, `tests/unit/test_ed25519_contract.c:21-22` | Three places in committed source assert the guarantee finding 1 shows is not implemented — most explicitly the generator docstring: "if either copy is edited, that side's build fails and the digests visibly stop matching". I edited a copy; nothing failed and the digest did not move. This is the same three-places-claim-it/nowhere-implements-it pattern as the previous round. | Fix finding 1 and the wording becomes true. If finding 1 is deferred, narrow the wording to what holds: the digest detects a *regenerated* corpus, not a hand-edited one. |

### What I verified

**Prior findings, all three closed.** `tests/vectors/ed25519_contract_vectors.h` is committed (422 lines, in `git ls-tree` at this head). `tests/CMakeLists.txt` has no `find_package(Python3 … REQUIRED)` and no `add_custom_command`. `git log origin/master..HEAD` is three commits touching three files — the FDT work from #84/#85 is gone. The Valgrind `foreach` includes `test_ed25519_contract`.

**Build and suite.** Clean build, 0 errors. `ctest --no-tests=error` → `100% tests passed, 0 tests failed out of 22`. The contract test reports `76` vectors, `3 accepted, 73 refused, 0 wrong`.

**The corpus is reproducible.** `python3 tools/gen_ed25519_contract_vectors.py` output is byte-identical to the committed header (`diff -q`, no output).

**The digest is honest.** I parsed the 76 vectors out of the committed header and recomputed SHA-256 over `pk ‖ sig ‖ msg ‖ expect_accept` myself: `1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2`, matching the pin, with 76 vectors and exactly 3 accepts. The data is what it claims to be — only the checking of it is loose.

**The eos twin really is byte-identical.** `git show refs/pr/122:tests/vectors/ed25519_contract_vectors.h` diffed against this branch's copy: no difference. The central claim of the PR holds today.

**Finding 1, reproduced.** One byte changed in the committed header, `0x01 → 0x02` in the first vector's public key, so it is no longer the identity point:

```
$ cmake --build build --target eboot_test_ed25519_contract && ./build/tests/eboot_test_ed25519_contract
Ed25519 contract vectors
  digest: 1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2
  count:  76
  3 accepted, 73 refused, 0 wrong
[PASS] all 76 contract vectors behaved as specified (3 must accept)
exit=0
```

The corpus silently lost its order-1 coverage and the digest never moved.

**The eight-messages-per-key design is right, and now measured.** This is the part of the PR I most wanted to check independently, because it is the argument the corpus size rests on. For each low-order key I computed the point's order and, for each of its eight messages, whether `order(A)` divides `k = SHA-512(R‖A‖M) mod L` — the condition under which `(R = identity, S = 0)` verifies against a verifier with no subgroup check:

| key | order | canonical | forgeries that land |
|---|---|---|---|
| `loworder_identity_order_1` | 1 | yes | 8/8 |
| `loworder_order_2` | 2 | yes | 5/8 |
| `loworder_noncanonical_ec_ff` | 2 | (x=0 with sign bit) | 6/8 |
| `loworder_order_4_zero` | 4 | yes | 1/8 |
| `loworder_order_4_signbit` | 4 | yes | 2/8 |
| `loworder_noncanonical_ed_ff` | 4 | no (y = p) | 4/8 |
| `loworder_order_8_a` | 8 | yes | 1/8 |
| `loworder_order_8_b` | 8 | yes | 1/8 |

The rates track `1/n` as the body predicts. The consequence is concrete: **a test pinned to one message would have missed both order-8 keys and `loworder_order_4_zero` with probability 7/8 each.** That is the justification for 64 vectors, and it is a real one.

It also reconciles the body's "28 of 76 wrong": 24 of those land on canonical keys, plus 4 on `loworder_noncanonical_ed_ff`. Worth knowing that the last 4 only count if the baseline verifier did not enforce y < p in `unpackneg()` — with canonicality enforced the figure is 24. Not a defect; the number is right for the baseline it was measured against, and it is worth writing down which baseline that was.

### Architecture conformance

Master design §5.1 and §8: **conforms.** Test-only change plus one tool; no runtime dependency in any direction, nothing added to `eboot_core`. eBoot is Tier 1 – Foundation (§21), correct repo.

On §21.1, the split policy, since this PR deliberately duplicates a file across two repositories: the corpus is not a subsystem seeking a repo, it is shared test data, and duplicating 54 KB of it with a pinned digest is a reasonable answer to a problem the design does not otherwise address. The master design has no concept of an artefact shared between repos — §10's component model covers runtime components, not test corpora — so there is no third home for this. Duplication plus a checked digest is the right call; finding 1 is only about the "checked" part.

`.ai/security.md` "Cryptography — reject invalid and low-order public keys explicitly; do not assume the library did", and "Key management is integrated across eBoot, eSec, eOTA and release signing (§14.1) — review it as one system, not four". This PR is the first thing in the tree that treats the two verifiers as one system, which is exactly the right instinct. Finding 1 is what stands between the instinct and the guarantee.

### Proposed changes

1. Replace the literal-to-literal `strcmp` with a runtime SHA-256 over the vector array, as specified in finding 1. Roughly fifteen lines, uses an API the file already includes, and provable: after the change, the one-byte tamper above must fail.
2. Make the identical change in eos#122 — the digest only means something if both sides compute it.
3. Narrow the three prose claims, or leave them once step 1 lands.

Step 1 is small and verifiable, but it is a change to a security test's semantics rather than a mechanical fix, so I have not opened a PR for it.

### Not checked

- **The "28 of 76 wrong against unfixed origin/master" run.** I could not reproduce it: `origin/master` does not compile (see #94), so there is no unfixed binary to run the corpus against. I verified the number analytically instead, which is a different kind of evidence — it confirms the arithmetic, not that the author ran what they say they ran.
- **The eos side beyond the two files I read.** I diffed eos#122's corpus header and read its driver's digest logic. I did not build eos, run its suite, or check its 35/35 claim.
- **Whether eBoot's `unpackneg()` implements RFC 8032 §5.1.3's `x = 0 && sign = 1` rejection.** It matters for *why* `loworder_noncanonical_ec_ff` is refused, not whether it is — the vector expects refusal and gets it. My own decoder does not implement that rule, which is why my table marks that key's order as 2 rather than treating it as undecodable.
- **Host x86-64 only.** The cross-compile and EoSim legs in `checks.txt` are green; not reproduced.
- **Valgrind.** `test_ed25519_contract` is registered in the Valgrind `foreach`, but I did not run the Valgrind targets.
- **Whether the three RFC 8032 §7.1 positive vectors are the real ones.** They are accepted by eBoot's verifier, which is evidence they are valid signatures under those keys; I did not check them against the RFC text itself.


---
*Automated architecture review of `ea9bc23809df` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (commented):
## Review — eBoot#89 "test(ed25519): hold both repos' verifiers to one shared contract"
head: 64a7f63d6c5cd81a636e8f90bea4c1e853cc9eff  author: Kartikey1306  ci: pass

**Verdict:** The design is right and it works — I built it, ran it, and regenerated the
corpus byte-for-byte. Recomputing the digest over the vector bytes rather than comparing two
literals from the same generated file is the detail that makes this a real guard, and the
three accept vectors plus the pinned accept-count close the refuse-everything hole. One
substantive problem: the corpus does not contain what it says it contains. Two of the eight
low-order encodings are missing, and one of the two entries named `noncanonical_*` is
canonical.

### Findings
|---|---|---|---|---|
| 1 | Medium | `tools/gen_ed25519_contract_vectors.py:33-46` | The list is labelled "The eight points of order dividing 8, plus the two non-canonical `y >= p` encodings" — that describes ten entries; there are eight, and only six are canonical low-order encodings. Computed each one (decode y, recover x, add until identity): orders `1, 2, 4, 4, 8, 8` for the first six. The two sign-flipped order-8 encodings — `26E8…6D53FC85` and `C717…92AC03FA`, both canonical, both order 8 — are absent. `tests/unit/test_ed25519.c:354-359` **in this same PR** lists them as part of "the eight", with a comment arguing that "a claimed class has to be the class it claims". The same argument applies here, and harder: this corpus is the cross-repo contract, so a verifier in either repo that handles the sign-bit variants differently passes it. The PR body's table inherits the error ("64 \| low-order keys — the eight points of order dividing 8"). | Add the two sign-flipped order-8 encodings to `LOW_ORDER`, which takes the corpus to 80 vectors, then regenerate, re-pin `EOS_ED25519_CONTRACT_EXPECTED`, `EOS_ED25519_CONTRACT_EXPECTED_COUNT` and the digest in `test_ed25519_contract.c` — **and in the eos twin (#122) in the same breath**, which is exactly the coordination this corpus exists to force. Fix the comment and the body table to say what the list holds. |
| 2 | Medium | `tools/gen_ed25519_contract_vectors.py:86-89` | The comment in `main()` says "The vectors are generated into the build tree rather than committed: this file is the only copy either repo carries, and the header it emits is 54 KB of generated data that used to be checked in to both." The opposite is true in this PR: `tests/vectors/ed25519_contract_vectors.h` is committed (424 lines), and `tests/CMakeLists.txt:88-97` argues at length for exactly that ("committed, not generated at build time… What the TCB's signature verifier was tested against belongs in the repository"). Two comments in one PR asserting opposite facts about the same file; the next maintainer to read the generator will delete the committed header. | Delete the stale paragraph. The CMakeLists comment is the one that matches the code. |
| 3 | Low | `tools/gen_ed25519_contract_vectors.py:44` | `noncanonical_ec_ff` (`ECFF…FFFF`) is not non-canonical. Masking the sign bit gives `y = 0x7FFF…FFEC = p − 1`, which is `< p`, so the encoding is canonical; it decodes to an order-2 point. It is the sign-flip of `order_2`, and a useful vector — just not the thing its name says. Only `noncanonical_ed_ff` (`y = p`) is genuinely non-canonical. Verified: `canonical=True, order=2` vs `canonical=False, order=4`. | Rename to `order_2_signbit`, and if a second non-canonical case is wanted, `y = p + 1` (reduces to the identity) is the natural partner — `tests/unit/test_ed25519.c:370-372` already uses it. |
| 4 | High | `core/keystore.c:20-36` | **Pre-existing on `master`, not introduced by this PR** — but this PR is what makes it checkable, so it belongs here. `default_dev_key` is documented as "the public half of TEST 1 in RFC 8032 section 7.1. The matching private key is printed in the RFC, so anyone at all can produce a signature that a bootloader trusting this key will accept." It is not that key. It is a mistranscription: 21 of 32 bytes match, then `…da a3 f4 a1 8c 42 c4 76 84 37 77 25` where RFC 8032 has `…da a6 23 25 af 02 1a 68 f7 07 51 1a`. Verified — the RFC test-1 signature is **refused** under the keystore constant and **accepted** under the real key. So the comment is wrong in both directions: nobody holds a private key for this value, and the "usable default for bring-up and for the unit tests" cannot verify anything. It fails closed, which is the safe direction, but a documented bring-up path that silently cannot work is worse than one that is absent, and a false claim about the TCB's trust anchor is the kind of thing `.ai/security.md` asks to be held strictly. | Separate PR, since it is not this diff. Two options: replace the constant with the real RFC 8032 test-1 public key (`d75a9801…f707511a`) and keep the `#warning`, or keep an orphan key and rewrite the comment to say it is deliberately unusable. Either way this PR's corpus is the right place for the assertion that pins it: add a vector, or a check in `test_keystore.c`, that the trust anchor is the key its comment names. |
| 5 | Low | generated header, comment block | The emitted header carries a garbled sentence: "Divergence is change on one side that does not reach the other is visible in review as two different digests." Two sentences spliced. It lands in a generated file committed to two repositories. | `tools/gen_ed25519_contract_vectors.py:112-113` — "A change on one side that does not reach the other is visible in review as two different digests." |
| 6 | Low | `tools/gen_ed25519_contract_vectors.py:88-95` | The `-o/--output` branch is never used: the docstring, the CMakeLists comment and the PR body all document `> tests/vectors/…`. Untested argument handling in a tool whose output is a TCB test corpus. | Either drop the flag, or use it in the documented invocation so it is the exercised path. |

### Architecture conformance
Master design §14.1 (integrate key management across eBoot, eSec, eOTA and release signing;
use reviewed primitives), §28 (evidence model — a claim needs the evidence its state
requires), §5.1 (eBoot keeps the TCB minimal and auditable), §21 Tier 1/Tier 2 boundary.
**Conforms, and closes a gap the design implies but does not state.**

§14.1 requires key management to be *integrated* across eBoot and eSec, but the design never
says how two repositories carrying independent implementations of the same primitive stay in
step. This PR answers that with shared data rather than shared code, which is the only
option that does not create a cross-repo runtime dependency — eBoot cannot depend on eos
without inverting §5.1. Sharing the corpus and pinning a digest on each side keeps the
dependency at review time, where it belongs. No `#include`, link line or manifest entry in
this diff points up a tier.

```
cmake -B build/host -G Ninja -DEBLDR_BUILD_TESTS=ON && cmake --build build/host
ctest --output-on-failure --no-tests=error -j4   ->  100% passed, 0 failed, 22/22

./tests/eboot_test_ed25519_contract
  digest: 1059febedae2b3e3dfaa1ef5b419fb37ed7f0d53b377a3250b53b95a546282a2
  count:  76
  3 accepted, 73 refused, 0 wrong
  [PASS] all 76 contract vectors behaved as specified (3 must accept)

python3 tools/gen_ed25519_contract_vectors.py | diff - tests/vectors/ed25519_contract_vectors.h
  -> identical.  stderr: vectors: 76  accept: 3  reject: 73
                         digest: 1059febe…46282a2
```

All three RFC 8032 §7.1 entries check out — I derived each public key from the RFC secret key
and reproduced each published signature, so the accept vectors really are the RFC's and not a
self-consistent substitute. That matters more than usual here, because those three carry the
entire weight of proving the verifier still accepts anything, and finding 4 shows this repo
has already mistranscribed one of these constants once.

The point worth keeping from the existing comment thread: the corpus found real signature
malleability in the eos verifier (`S` in `[L, 2^253)` accepted) that eBoot's
`s_is_canonical()` already covered. That is the guard working before it was even merged, and
it is the argument for §14.1 being read as one system rather than four.

### Proposed changes
1. Add the two sign-flipped order-8 encodings; regenerate; re-pin count and digest here and
   in eos#122 together (finding 1).
2. Delete the contradictory paragraph in `main()` (finding 2).
3. Rename `noncanonical_ec_ff` → `order_2_signbit`; optionally add `y = p + 1` (finding 3).
4. Fix the spliced sentence (finding 5). Findings 1, 3 and 5 all land in one regeneration.
5. Open a separate PR for `core/keystore.c` (finding 4) and add the pinning assertion here.
6. The PR body's table needs the same correction as finding 1.

### Not checked
- Host `x86_64` `gcc` only. No cross-compile, no on-hardware run, no valgrind pass —
  `test_ed25519_contract` was added to the valgrind list at `tests/CMakeLists.txt:136` and
  that list was not exercised.
- The body's "against unfixed `origin/master` -> 28 of 76 wrong, exit 1" was **not**
  reproduced. `origin/master@22d8f8b` does not compile (duplicate `point_is_identity` in
  `core/ed25519_verify.c`, and `include/eos_image.h` asserting on a removed `reserved`
  member), so there is no buildable "unfixed master" to run the corpus against. Which commit
  that measurement was taken on is unclear; #86 is named as the dependency but is not in this
  bundle.
- The eos twin (embeddedos-org/eos#122) was not inspected. Whether its copy of the header is
  byte-identical and pins the same digest — the property the whole design rests on — is
  unverified from here. The malleability claim in the comment thread is likewise unverified.
- No fuzzing. `.ai/security.md` asks for fuzz coverage on externally reachable parsers; 76
  fixed vectors are a contract, not a fuzz corpus, and this PR does not claim otherwise.
- `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED` — the branch protection rule
  behind that was not inspected.
  `tests/unit/test_ed25519.c` changes that repair `master`'s broken host build, identical to
  eBoot#81 and #88. Reviewed under #81; not re-litigated here.


---
*Automated architecture review of `64a7f63d6c5c` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #94, opened 2026-09-03
fix: repair master — the ABI asserts and the Ed25519 verifier both merged broken

Kartikey1306:
`master` (22d8f8b) does not compile. Two independent double-merges, both the same shape: two PRs fixing adjacent things landed on stale bases, each green on its own branch, and the merged result was never rebuilt.

```
$ cmake -S . -B build -DEBLDR_BUILD_TESTS=ON && cmake --build build
include/eos_image.h:135:23: error: no member named 'reserved' in 'eos_image_header_t'
include/eos_image.h:142:57: error: no member named 'reserved' in 'eos_image_header_t'
```

This blocks every open PR that builds the test suite.

## 1. `include/eos_image.h` — #93 and #87

#93 replaced `reserved[30]` with `tlv_len` (2 bytes) + `tlv_hash[28]`, preserving every offset. #87 merged afterwards carrying asserts written against the older struct.

#93 already asserts `tlv_len` at 62 and `tlv_hash` at 64, so #87's offset assert was a duplicate. Its **width** assert had no replacement, and that is the one that matters — an offset assert cannot see a field growing into padding that keeps every later offset. Restored as two asserts covering both halves of the same 30-byte span. No offset moves; the wire format is unchanged.

## 2. `core/ed25519_verify.c` — #86 and #57

Both landed a subgroup guard, so the file carried two byte-identical `point_is_identity()` definitions. Only #57's `public_key_is_valid_subgroup()` is wired to the call site, so #86's `key_has_prime_order()` was dead code. Kept the live function, folded #86's fuller rationale onto it, deleted the duplicate.

## 3. `tests/unit/test_ed25519.c` — collateral from the same merge

Two copies of `test_ed25519_identity_key_forgery_rejected`; `main()` calling it twice and two tests not at all; and `test_ed25519_low_order_R_with_a_valid_key_is_not_a_forgery` referencing `k_low_order[]` and `messages[]` that the merge had dropped.

**While restoring the corpus, corrected it** — this is the review finding on #86. The array claimed to hold "the eight low-order point encodings" and held five. Every order below was *computed*, not copied: decode `y`, recover `x`, add the point to itself until it reaches the identity.

| encoding | order | in the old array? |
|---|---|---|
| `01 00…00` | 1 (identity) | yes |
| `ECFF…FF7F` | 2 | yes |
| `00…00` | 4 | yes |
| `00…0080` | 4 | **no** |
| `26E8…6D53FC05` | 8 | yes |
| `C717…92AC037A` | 8 | yes |
| `26E8…6D53FC85` | 8 | **no** |
| `C717…92AC03FA` | 8 | **no** |
| `D9FF…FF` | *does not decode* | yes — and it is not a low-order point |

`D9FF…FF` moves to a separate `k_non_canonical[]` alongside `EDFF…FF7F` (`y = p`) and `EEFF…FF7F` (`y = p+1`). `[L](-A) = -[L]A` so the guard rejects a sign variant whether or not it is listed — the reason to list them is that this is the regression record for a Critical secure-boot bypass, and a claimed class has to be the class it claims.

`tests_run` was assigned a literal (`11`) in `main()` and never incremented, which is exactly how a duplicate call and two unregistered tests went unnoticed. The `TEST` macro now increments it, so the total cannot drift from what actually ran.

## Verification

| check | before | after |
|---|---|---|
| `cmake --build` (`EBLDR_BUILD_TESTS=ON`) | **3 errors** | clean |
| `ctest` | could not build | **21/21 PASS** |
| `ctest` under `-DEBLDR_SANITIZE=ON` (ASan+UBSan) | — | **21/21 PASS** |
| `pytest tests/` | — | 24 passed, 1 skipped |
| `test_ed25519` | 11 claimed, 12 run, 1 twice | **14/14 PASS** |

The sweep still discriminates. With `public_key_is_valid_subgroup()` disabled:

```
test_ed25519_low_order_keys_rejected            [FAIL]   <- as it must
test_ed25519_non_canonical_encodings_rejected   [PASS]   <- different mechanism
```

That second line is why the two arrays are separate: the non-canonical encodings are refused by `unpackneg()` on canonicality, not by the subgroup check, so nobody should delete that path believing the subgroup test now covers it.

Refs #87, #93, #86, #57

### DIFF
--- a/core/ed25519_verify.c
+++ b/core/ed25519_verify.c
@@ -289,6 +289,16 @@ static int point_is_identity(gf p[4])
     return diff == 0;
 }
 
+static void scalarbase(gf r[4], const uint8_t *s)
+{
+    gf q[4];
+    fe_copy16(q[0], BX);
+    fe_copy16(q[1], BY);
+    fe_copy16(q[2], gf1);
+    fe_mul(q[3], BX, BY);
+    scalarmult(r, q, s);
+}
+
 /* Reject a public key outside the prime-order subgroup.
  *
  * Decoding a point is not enough. Ed25519 has eight points of low order, and
@@ -304,51 +314,9 @@ static int point_is_identity(gf p[4])
  * so there is no separate constant to transcribe wrongly: a mistyped L would
  * reject valid keys, and only in the field.
  *
- * A arrives negated from unpackneg(). [L](-A) = -[L]A and the identity is its
- * own negation, so neither condition is affected by the sign.
- *
- * Formulation taken from eBoot#57 by @muhammadburhandevv-hub, which reached
- * this before I did and states both conditions in one expression.
+ * The key arrives negated from unpackneg(). [L](-A) = -[L]A and the identity
+ * is its own negation, so neither condition is affected by the sign.
  */
-static int key_has_prime_order(gf A[4])
-{
-    uint8_t order_l[32];
-    gf q[4], multiple[4];
-    int i;
-
-    for (i = 0; i < 32; i++)
-        order_l[i] = (uint8_t)ORDER_L[i];
-    for (i = 0; i < 4; i++)
-        fe_copy16(q[i], A[i]);
-
-    scalarmult(multiple, q, order_l);
-    return point_is_identity(multiple) && !point_is_identity(A);
-}
-
-static void scalarbase(gf r[4], const uint8_t *s)
-{
-    gf q[4];
-    fe_copy16(q[0], BX);
-    fe_copy16(q[1], BY);
-    fe_copy16(q[2], gf1);
-    fe_mul(q[3], BX, BY);
-    scalarmult(r, q, s);
-}
-
-static int point_is_identity(gf p[4])
-{
-    uint8_t encoded[32];
-    point_pack(encoded, p);
-
-    uint8_t diff = (uint8_t)(encoded[0] ^ 1U);
-    for (int i = 1; i < 32; i++)
-        diff |= encoded[i];
-    return diff == 0;
-}
-
-/* Public keys must be non-identity points in Ed25519's prime-order subgroup.
- * Merely decoding a point is insufficient: an identity or torsion key can
- * make the verification equation true without knowledge of a private key. */
 static int public_key_is_valid_subgroup(gf public_key[4])
 {
     uint8_t order_l[32];
--- a/include/eos_image.h
+++ b/include/eos_image.h
@@ -108,7 +108,8 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, tlv_hash) +
 
 /* Every remaining field, pinned.
  *
- * Four of the fourteen fields were asserted. Transposing two adjacent
+ * Three of the thirteen field offsets were asserted (the fourth pre-existing
+ * assert is sizeof, which is not a field). Transposing two adjacent
  * same-width fields moves neither sizeof nor any of those four offsets, so it
  * compiled clean: with load_addr and entry_addr swapped, all four existing
  * asserts still passed and the bootloader would load an image at its entry
@@ -132,15 +133,18 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, flags) == 24,
                       "flags must stay at offset 24");
 EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, sig_len) == 61,
                       "sig_len must stay at offset 61");
-EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, reserved) == 62,
-                      "reserved[] must stay at offset 62");
+/* tlv_len and tlv_hash are asserted above, where #93 introduced them; the
+ * 30 bytes they occupy are the ones this block used to pin as reserved[]. */
 
 /* Field widths. An offset assert cannot see a field growing into padding that
- * happens to keep every later offset -- reserved[] absorbs exactly that. */
+ * happens to keep every later offset -- the 30 bytes at 62 absorb exactly
+ * that, which is why both halves of that span carry a width assert. */
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->hash) == 32,
                       "hash[] is 32 bytes on the wire");
-EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->reserved) == 30,
-                      "reserved[] is 30 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_len) == 2,
+                      "tlv_len is 2 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_hash) == 28,
+                      "tlv_hash is 28 bytes on the wire");
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->signature) == 64,
                       "signature[] is 64 bytes on the wire");
 
--- a/tests/unit/test_ed25519.c
+++ b/tests/unit/test_ed25519.c
@@ -30,6 +30,7 @@ static int tests_passed = 0;
     static void name(void); \
     static void run_##name(void) { \
         printf("  %-50s ", #name); \
+        tests_run++; \
         name(); \
         tests_passed++; \
         printf("[PASS]\n"); \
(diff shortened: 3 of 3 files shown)


srpatcha (commented):
## Review — eBoot#94 "fix: repair master — the ABI asserts and the Ed25519 verifier both merged broken"
head: f704d87fd445e4d61db00bc302e78660fc99f02f  author: Kartikey1306  ci: pass

**Verdict:** Correct, and every claim in the body checks out — I reproduced the breakage on `master`, reproduced the repair, and independently recomputed the low-order point orders rather than taking the table on trust. This is the base of the #84 → #85 and #89 stacks and blocks all of them. One finding, and it is about what the PR deliberately does not do.

### Findings

|---|---|---|---|---|
| 1 | Medium | org branch protection, not a file in this diff | The body diagnoses the cause precisely — "two PRs fixing adjacent things landed on stale bases, each green on its own branch, and the merged result was never rebuilt" — and then repairs only the damage. Nothing here stops the third occurrence, and this PR is already fixing two independent instances of it (#87/#93 and #86/#57). Every required check in `checks.txt` runs against the PR head; none runs against the merge result, so two green PRs can and did produce a red `master`. | Not a code change: enable **Require branches to be up to date before merging** on `master` for the required checks, or turn on a merge queue. Worth stating in the PR so it is not lost with the fix. `.github/STANDARDS.md` §"Release model" defines the branch model but says nothing about merge preconditions — see the proposal appended to `.ai/autoreview/proposals/2026-09.md`. |

Nothing else. The diff does what it says, no check is weakened, nothing is deleted without explanation, and the two files it touches are the two that are broken.

### What I verified

**The breakage is real.** `origin/master` at `22d8f8b`:

```
$ cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DEBLDR_BUILD_TESTS=ON && cmake --build build
gmake: *** [Makefile:101: all] Error 2

$ gcc -fsyntax-only -Iinclude core/ed25519_verify.c
```

**The repair is complete.** At this head, clean build and:

```
100% tests passed, 0 tests failed out of 21
14/14 tests passed          # test_ed25519
```

Both numbers match the body exactly.

**`key_has_prime_order()` really was dead.** On master the only call site is line 496, `if (!public_key_is_valid_subgroup(A))`; `key_has_prime_order` is defined at 313 and never referenced. Deleting the right one of the two.

**The retained guard fails closed and does both halves.** `public_key_is_valid_subgroup()` returns `point_is_identity(multiple) && !point_is_identity(public_key)`, and the caller maps a false to `EOS_ERR_SIGNATURE`. `ORDER_L` is the real group order — I decoded the 32 bytes at `core/ed25519_verify.c:71` and compared against `2^252 + 27742317777372353535851937790883648493`: equal.

**I recomputed the low-order table** rather than trusting it, since it is the regression record for a secure-boot bypass. Decoding each `y`, recovering `x`, and repeatedly adding the point to itself until it reaches the identity:

```
k_low_order[]                            k_non_canonical[]
order 1 identity   order=1  canonical    y = p      decodes, order 4, NOT canonical
order 2            order=2  canonical    y = p+1    decodes, order 1, NOT canonical
order 4 sign clear order=4  canonical    D9FF..FF   no x exists on the curve
order 4 sign set   order=4  canonical
order 8 a          order=8  canonical
order 8 b          order=8  canonical
order 8 a'         order=8  canonical
order 8 b'         order=8  canonical
```

`1, 2, 4, 4, 8, 8, 8, 8` is the whole 8-element torsion subgroup, so the array is now the class it names, and the split is justified on exactly the grounds the body gives: `D9FF..FF` has no `x` at all and was never a low-order point, while `y = p` and `y = p+1` decode but are non-canonical and are refused earlier by `unpackneg()`. Every claim in the body's table holds.

**One thing the body understates.** It says `tests_run` "was assigned a literal (`11`) and never incremented, which is exactly how a duplicate call and two unregistered tests went unnoticed". It was worse than mis-labelling: master's `main()` makes 13 `run_*` calls (12 distinct, `identity_key_forgery` twice), so `tests_passed` reaches 13 against `tests_run = 11` and the binary **exits 1**. And `test_ed25519_low_order_R_with_a_valid_key_is_not_a_forgery` is defined at `tests/unit/test_ed25519.c:278` on master with zero call sites — a signature-forgery test that never ran. `git log -S "tests_run = 11"` puts both in `22d8f8b`, the same merge, so the exposure is recent rather than long-standing. This PR restores the call and makes the count self-maintaining; that is a weakened check being repaired and belongs in the body.

### Architecture conformance

Master design §5.1 and §8: **conforms.** No dependency changes, no new files, no tier movement. eBoot, Tier 1 – Foundation (§21). The change is confined to restoring the state the two merges destroyed.

`.ai/security.md` "Fail closed" and "Cryptography — reject invalid and low-order public keys explicitly": this PR is what puts eBoot back on the right side of both. Master today cannot build, so `Host Build & Tests` on every open PR is testing the PR's own repair rather than the tree.

The design-level gap is §28, "Status, Evidence and Claims Policy". It defines what evidence an *artifact* needs to claim a state — "Implemented: code and functional tests" — and is silent on the integration gate that keeps the trunk in that state. An org whose `master` does not compile still satisfies §28 as written, because §28 never asks the question of the merged result. That is a hole in the master design rather than in this PR; proposal appended to `.ai/autoreview/proposals/2026-09.md`.

### Proposed changes

1. Land this first. #84, #89 and (through #84) #85 are all based on `f704d87` and cannot build without it.
2. Add finding 1 to the body as a follow-up item, or open an issue, so the branch-protection change is not lost once the tree is green again.
3. Add the two facts above — `test_ed25519` exiting 1, and the uncalled forgery test — to §3 of the body. They make the case for the `tests_run` change stronger than it currently reads.

No code change requested.

### Not checked

- **`pytest tests/` — "24 passed, 1 skipped".** I did not run it; I ran the CMake/ctest path only. Unverified.
- **The disabled-guard experiment.** The body shows `low_order_keys_rejected [FAIL]` / `non_canonical_encodings_rejected [PASS]` with `public_key_is_valid_subgroup()` disabled. I did not reproduce it. It is consistent with my own computation — the non-canonical entries are refused by `unpackneg()` before the subgroup check runs — but consistency is not the same as having run it.
- **Sanitizer run at this head.** Body claims 21/21 under `-DEBLDR_SANITIZE=ON`. I ran ASan/UBSan on #84 (22/22, this commit plus the FDT work) and not separately here.
- **Host x86-64 only.** The eight cross-compile and eleven EoSim jobs in `checks.txt` are green; not reproduced.
- **Why `Host Build & Tests` was green on `master`'s own merge commits.** If `test_ed25519` was exiting 1 from `22d8f8b` onward, something should have been red. I did not dig into the workflow definitions to find out whether the job runs `ctest` at all on push, so I cannot tell you whether the signal was missing or merely ignored. Worth someone checking — it is the same question finding 1 asks.


---
*Automated architecture review of `f704d87fd445` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
Worth flagging for merge priority: **this blocks every open PR in `eos` as well as in this repository.**

`eos/.github/workflows/eos-simulation.yml:28` checks out `embeddedos-org/eBoot` at `ref: master` and builds it. So the `EoS Full-Stack Simulation` job fails identically on all nine open eos PRs, none of which touch anything related:

```
Cross-compile ARM64 kernel  Build eBoot (ARM64 qemu_arm64 board)
  eBoot/include/eos_image.h:135:23: error: 'eos_image_header_t' has no member named 'reserved'
  eBoot/include/eos_image.h:142:55: error: 'eos_image_header_t' has no member named 'reserved'
```

That is this PR's finding 1, reached from the other repository. It is the only failing check on eos #114, #116, #117, #118, #119, #122, #126 and #127 — every other job on all of them is green — and nothing can be done from the eos side, because the workflow builds `master`, not the PR.

The eBoot side is the same story: #80, #82, #84, #85, #88, #89 and #90 are all stacked on this branch because they cannot build otherwise.

No review shortcut intended — the diff is three files and the evidence is in the body. Just noting that the blast radius is larger than the PR looks.

Kartikey1306:
This PR is blocking **11 open pull requests in `eos`**, and it is the single
highest-leverage merge in the three repos right now. Evidence, since that is a
claim about other repositories.

## The blast radius

`eos`'s `.github/workflows/eos-simulation.yml` checks out
`embeddedos-org/eBoot@master` and cross-builds it (`Cross-compile ARM64
kernel`, `qemu_arm64`). `eBoot` master does not compile, so that job fails on
**every** `eos` pull request, and `Full-stack integration summary` fails
downstream of it.

Swept every open `eos` PR head just now. `EoS Full-Stack Simulation` is red on:

```
#114 #115 #116 #117 #118 #119 #121 #122 #126 #127 #129
```

— eleven, all with the identical signature:

```
eBoot/include/eos_image.h:135: error: 'eos_image_header_t' has no member named 'reserved'
eBoot/include/eos_image.h:142: error: ...
```

None of those PRs touches `eBoot`. Every one of them is carrying a red X that
belongs here.

## Reproduced locally, both directions

The failure is a `_Static_assert` in a header, so it needs no ARM toolchain —
any compiler that parses `eos_image.h` hits it. Compiling a two-line
translation unit that does nothing but include the header:

```
eBoot master : FAILS
    eos_image.h:135:23: error: no member named 'reserved' in 'eos_image_header_t'
    eos_image.h:142:57: error: no member named 'reserved' in 'eos_image_header_t'
    2 errors generated.
eBoot #94    : COMPILES CLEAN
```

So the fix in this PR is exactly the fix for the eos-side failure, and I can
confirm that without waiting for CI to tell us.

## Why it happened, for the record

`master` asserts `offsetof(eos_image_header_t, reserved) == 62` against a struct
that no longer has a `reserved` member — the TLV work replaced those 30 bytes
and the assert pinning them was left behind. A static assert on a field that
does not exist is not a weakened check; it is a header that cannot be included
at all, which is why it takes out every consumer rather than one test.

## Ask

Nothing to change here — this PR is `MERGEABLE`, its own checks are green
(4 pass, 2 skipped), and I have verified it fixes the defect. **Landing it
clears one red X from eleven `eos` PRs in one merge**, which is a better return
than anything else currently queued.

If it helps triage: the only *other* recurring red on those eos PRs is the
`test_crypto_ed25519_loworder` registration, which #127 in eos fixes. Between
this and eos#127, the eos queue goes green.

Kartikey1306:
Root cause identified, and it is my own merged PR. Owning it plainly since it
changes how urgent this is.

## Which commit broke `master`, and how

```
abd4dab  Fix/authenticate tlv anti rollback (#93)      <- removed uint8_t reserved[30]
bbd997a  fix(image): pin the whole .efw header (#87)   <- added offsetof(..., reserved) asserts
```

Bisecting the header across master's recent commits:

```
v3.0.1    eos_image.h errors: 0
abd4dab   eos_image.h errors: 0
bbd997a   eos_image.h errors: 2   <- breaks here
32723b3   eos_image.h errors: 2
22d8f8b   eos_image.h errors: 2   (current master)
```

#93 replaced `reserved[30]` with the TLV fields. **#87 — mine — was written
against a base where `reserved` still existed, merged after #93, and added**

```c
EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, reserved) == 62, ...);
EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->reserved) == 30, ...);
```

against a member that was no longer there. Both PRs were individually green;
neither was rebuilt against the other before landing. That is the exact
failure mode #92 and #87-the-issue describe, and I produced an instance of it
while arguing for the gate that would have caught it.

## Confirmed this PR is the complete fix

Full build, not just the header:

```
eBoot #94   : configure OK, build errors 0, ctest 21/21
eBoot master: build errors 10
```

## Why it is worth merging ahead of the queue

`eos`'s `eos-simulation.yml` checks out `embeddedos-org/eBoot@master` and
cross-builds it, so this breakage fails `Cross-compile ARM64 kernel` on **every
open eos pull request**. Currently red for this reason:

```
eos #114 #115 #116 #117 #118 #119 #121 #122 #127 #129 #134
```

Eleven PRs, none of which touch eBoot. Merging this clears all eleven.

## Separate observation, not a proposal for this PR

That coupling is worth revisiting on its own: pointing PR CI at another
repository's moving default branch means their breakage blocks your merges and
you cannot fix it from your side. Pinning the integration build to a release
tag for pull requests, and keeping a scheduled run against `master` so drift is
still caught loudly, would decouple them without going fail-open. I am
deliberately **not** folding that into this PR — it would look like routing
around my own breakage rather than fixing it, and this PR should stay the
minimal repair.

Kartikey1306:
Evidence for prioritizing this: merging it takes **12 currently-red eos PRs green in one step**, with no change needed on any of them.

Every open eos PR fails exactly one job — `EoS Full-Stack Simulation` — because eos's workflow builds `eBoot@master`, and master does not compile. Verified locally rather than inferred, under that job's own configuration (`-DEBLDR_BOARD=qemu_arm64 -DEBLDR_VERIFY_STAGE1=OFF -DCMAKE_BUILD_TYPE=Release`, cross-compile mode):

| ref | result |
|---|---|
| `master` (22d8f8b) | **fails** — `eos_image.h:135/:142`: `offsetof(eos_image_header_t, reserved)` for a member that became `tlv_len` + `tlv_hash` |
| this PR's head | **builds every target, rc=0**, and its suite is 21/21 |

Master carries a second, independent break this PR also fixes: `point_is_identity` redefined in `core/ed25519_verify.c` — it entered when #57 merged on top of #86's near-identical hardening. Walking master back and *building* each candidate, the newest compiling ancestor is `a172a6d` (its child `8a015b2` already fails).

Both defects are pairs of PRs that merged clean and broke the build together; eos PR #135 pins eos's eBoot checkout to `a172a6d` so the next such pair cannot redden every eos PR again, and its own simulation run is green on that pin. Once this merges, that pin bumps to this merge SHA.

Since eos PR CI runs on the merge commit with master, the 12 eos PRs go green on their next run after this lands — no rebases, no pushes.

PR #99, opened 2026-09-03
fix(fuzz): point every harness at the API it claims to fuzz

Kartikey1306:
> **Stacked on #94**, without which `include/eos_image.h` does not compile.

Four of the five fuzz harnesses did not test what they name, and nothing in the build could say so, because **not one of them includes a project header**. Each declares its own target:

```c
fuzz_recovery_protocol.c  extern int eos_recovery_parse_packet(const uint8_t *, size_t);
fuzz_fw_update.c          extern int eos_fw_update_init(void);
                          extern int eos_fw_update_process_chunk(const uint8_t *, size_t);
                          extern int eos_fw_update_finalize(void);
fuzz_bootctl.c            extern int eos_bootctl_parse(const void *, size_t);
fuzz_image_verify.c       extern int eos_image_parse_header(const void *, size_t);
```

`eos_recovery_parse_packet`, `eos_fw_update_init`, `eos_fw_update_process_chunk` and `eos_bootctl_parse` **have never existed** — each name appears nowhere outside the harness declaring it. Those three targets have never linked, which is what fails `Fuzz Harness Build` on #85.

## The two that do link are worse

- `eos_fw_update_finalize` exists, but takes `(ctx, mode)` — not `(void)`.
- `eos_image_parse_header` takes `(uint32_t addr, eos_image_header_t *out)` — not `(const void *, size_t)`. Same name, different function; C never checks across translation units.

`fuzz_image_verify` has therefore been passing the fuzzer's pointer as a flash address and its length as an output struct, on every input, for as long as it has run. Measured over 20,000 inputs:

| parser result | before | after |
|---|---|---|
| `EOS_ERR_FLASH` (-6) — first flash read failed | **20000** | 0 |
| `EOS_ERR_NO_IMAGE` (-5) — magic check reached | 0 | **10000** |
| `EOS_ERR_INVALID` (-2) — version/hdr_size reached | 0 | **10000** |

Every single input bailed two lines in. The target ran, reported nothing, and covered none of the parser.

## What each harness drives now

| harness | real entry points |
|---|---|
| `fuzz_image_verify` | `parse_header` → `verify_integrity`, in boot-path order |
| `fuzz_bootctl` | `bootctl_load`, then trap if `load()` accepted a block `validate()` rejects |
| `fuzz_fw_update` | `begin` → `write` in fuzz-chosen chunk widths → `finalize`/`abort` |
| `fuzz_recovery_protocol` | `write_in_range`, trapping on any accepted write that leaves the slot or wraps — rechecked in 64-bit types that cannot |
| `fuzz_crypto` | already correct; moved to the header so it stays that way |

New `tests/fuzz/fuzz_sim_flash.h`: the three harnesses that reach flash need board ops installed, and without them they fuzz a null op table.

**That the compiler now checks this is not theoretical** — writing it, it rejected `EOS_UPGRADE_MODE_TEST` for `EOS_UPGRADE_TEST` immediately. That is precisely the class the `extern`s were hiding.

## Validation

libFuzzer is unavailable here (`libclang_rt.fuzzer_osx.a` absent), so each harness was linked against `eboot_core` + `eboot_hal` and driven by a standalone seed generator under `-fsanitize=address,undefined`:

```
compile -Wall -Wextra   5/5, 0 failures
20,000 inputs each      100,000 total, no reports
ctest                   21/21 passed
```

## Credit

**@task-13's session independently found and fixed the same three harnesses**, and reached the identical before/after conclusion on `fuzz_image_verify` — its parked branch is `fix/fuzz-harnesses-that-link`. It deliberately did not open a PR to avoid duplicating this one. This PR is the superset: it also covers `fuzz_bootctl` (whose `eos_bootctl_parse` is equally non-existent) and `fuzz_crypto`, so no `extern` is left in `tests/fuzz/`. Its deeper `fuzz_fw_update` body is worth layering on top of this if you want it.

## The pattern

Third instance across these repos, after eos#50 (`fuzz_devicetree` naming `eos_dtb_parse`) and eos#110 (`fuzz_ota_header` naming `eos_ota_parse_header`). The common factor every time is **a fuzz target declaring its own subject**. Removing the last `extern` from `tests/fuzz/` is what stops the fourth.

### DIFF
--- a/core/ed25519_verify.c
+++ b/core/ed25519_verify.c
@@ -289,6 +289,16 @@ static int point_is_identity(gf p[4])
     return diff == 0;
 }
 
+static void scalarbase(gf r[4], const uint8_t *s)
+{
+    gf q[4];
+    fe_copy16(q[0], BX);
+    fe_copy16(q[1], BY);
+    fe_copy16(q[2], gf1);
+    fe_mul(q[3], BX, BY);
+    scalarmult(r, q, s);
+}
+
 /* Reject a public key outside the prime-order subgroup.
  *
  * Decoding a point is not enough. Ed25519 has eight points of low order, and
@@ -304,51 +314,9 @@ static int point_is_identity(gf p[4])
  * so there is no separate constant to transcribe wrongly: a mistyped L would
  * reject valid keys, and only in the field.
  *
- * A arrives negated from unpackneg(). [L](-A) = -[L]A and the identity is its
- * own negation, so neither condition is affected by the sign.
- *
- * Formulation taken from eBoot#57 by @muhammadburhandevv-hub, which reached
- * this before I did and states both conditions in one expression.
+ * The key arrives negated from unpackneg(). [L](-A) = -[L]A and the identity
+ * is its own negation, so neither condition is affected by the sign.
  */
-static int key_has_prime_order(gf A[4])
-{
-    uint8_t order_l[32];
-    gf q[4], multiple[4];
-    int i;
-
-    for (i = 0; i < 32; i++)
-        order_l[i] = (uint8_t)ORDER_L[i];
-    for (i = 0; i < 4; i++)
-        fe_copy16(q[i], A[i]);
-
-    scalarmult(multiple, q, order_l);
-    return point_is_identity(multiple) && !point_is_identity(A);
-}
-
-static void scalarbase(gf r[4], const uint8_t *s)
-{
-    gf q[4];
-    fe_copy16(q[0], BX);
-    fe_copy16(q[1], BY);
-    fe_copy16(q[2], gf1);
-    fe_mul(q[3], BX, BY);
-    scalarmult(r, q, s);
-}
-
-static int point_is_identity(gf p[4])
-{
-    uint8_t encoded[32];
-    point_pack(encoded, p);
-
-    uint8_t diff = (uint8_t)(encoded[0] ^ 1U);
-    for (int i = 1; i < 32; i++)
-        diff |= encoded[i];
-    return diff == 0;
-}
-
-/* Public keys must be non-identity points in Ed25519's prime-order subgroup.
- * Merely decoding a point is insufficient: an identity or torsion key can
- * make the verification equation true without knowledge of a private key. */
 static int public_key_is_valid_subgroup(gf public_key[4])
 {
     uint8_t order_l[32];
--- a/include/eos_image.h
+++ b/include/eos_image.h
@@ -108,7 +108,8 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, tlv_hash) +
 
 /* Every remaining field, pinned.
  *
- * Four of the fourteen fields were asserted. Transposing two adjacent
+ * Three of the thirteen field offsets were asserted (the fourth pre-existing
+ * assert is sizeof, which is not a field). Transposing two adjacent
  * same-width fields moves neither sizeof nor any of those four offsets, so it
  * compiled clean: with load_addr and entry_addr swapped, all four existing
  * asserts still passed and the bootloader would load an image at its entry
@@ -132,15 +133,18 @@ EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, flags) == 24,
                       "flags must stay at offset 24");
 EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, sig_len) == 61,
                       "sig_len must stay at offset 61");
-EOS_IMG_STATIC_ASSERT(offsetof(eos_image_header_t, reserved) == 62,
-                      "reserved[] must stay at offset 62");
+/* tlv_len and tlv_hash are asserted above, where #93 introduced them; the
+ * 30 bytes they occupy are the ones this block used to pin as reserved[]. */
 
 /* Field widths. An offset assert cannot see a field growing into padding that
- * happens to keep every later offset -- reserved[] absorbs exactly that. */
+ * happens to keep every later offset -- the 30 bytes at 62 absorb exactly
+ * that, which is why both halves of that span carry a width assert. */
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->hash) == 32,
                       "hash[] is 32 bytes on the wire");
-EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->reserved) == 30,
-                      "reserved[] is 30 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_len) == 2,
+                      "tlv_len is 2 bytes on the wire");
+EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->tlv_hash) == 28,
+                      "tlv_hash is 28 bytes on the wire");
 EOS_IMG_STATIC_ASSERT(sizeof(((eos_image_header_t *)0)->signature) == 64,
                       "signature[] is 64 bytes on the wire");
 
--- a/tests/fuzz/fuzz_bootctl.c
+++ b/tests/fuzz/fuzz_bootctl.c
@@ -3,25 +3,43 @@
 
 /**
  * @file fuzz_bootctl.c
- * @brief libFuzzer harness for boot control block (BCB) parsing
+ * @brief libFuzzer harness for boot control block loading and validation.
  *
- * Feeds arbitrary data into the boot control block parser, exercising
- * magic-number validation, slot metadata decoding, retry counters,
- * and CRC integrity checks.
+ * This harness used to declare and call `eos_bootctl_parse()`. No such
+ * function exists in this repository -- the name appears nowhere outside this
+ * file -- so the target never linked, and a fuzz target that does not link
+ * proves nothing about the code it names.
+ *
+ * The real untrusted-input path is eos_bootctl_load(), which reads the block
+ * out of flash and is where a corrupt or hostile BCB arrives. The fuzz input
+ * is written to the primary and backup BCB addresses and loaded from there.
  */
 
-#include <stdint.h>
+#include "eos_bootctl.h"
+#include "fuzz_sim_flash.h"
+
 #include <stddef.h>
+#include <stdint.h>
 
-/* Forward-declare boot control block parser */
-extern int eos_bootctl_parse(const void *bcb_data, size_t bcb_len);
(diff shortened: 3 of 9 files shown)


Kartikey1306:
The `Fuzz Harness Build` job on #101 (stacked on this) ran these harnesses under
libFuzzer for the first time, and `fuzz_recovery_protocol` trapped in six
seconds. It is a false positive: the oracle at
`tests/fuzz/fuzz_recovery_protocol.c:45` is off by one, and
`eos_recovery_write_in_range()` is correct.

Crashing unit, from job `100752628399`:

```
0xdb,0xdb,0xff,0xff, 0xdb,0xdb,0x01,0x00, 0x00,0x00,0x00,0x00, 0x25,0x24
base=0xFFFFDBDB  slot_size=0x0001DBDB  offset=0  len=9253
```

Reproduced locally against `eboot_core`:

```
rc=0  end=0x100000000  last byte=0xFFFFFFFF
```

The write covers `[0xFFFFDBDB, 0xFFFFFFFF]` — 9253 bytes ending on the last
addressable byte. Every byte of it is inside 32 bits, so accepting it is right.

The oracle asserts on `base + offset + len > 0xFFFFFFFF`. That is the address
*one past* the end. The last byte written is `base + offset + len - 1`, so the
condition that actually has to hold is

```c
end - 1 > 0xFFFFFFFFULL          /* or, equivalently, end > 0x100000000ULL */
```

`recovery_handle_write()` calls `eos_hal_flash_write(base + offset, buf, len)`,
so `len - 1` is the right displacement, and the implementation already says so
deliberately in its last line:

```c
if ((uint32_t)len - 1u > UINT32_MAX - (base + offset))
    return EOS_ERR_INVALID;
```

So the harness is currently stricter than the contract, and it fails the fuzz
job on a legal input. One character, at `:45`:

```c
-            end > 0xFFFFFFFFULL) {
+            end - 1 > 0xFFFFFFFFULL) {
```

The other three clauses are right, and the choice to give this harness an oracle
rather than only watching for crashes is the right one — a predicate that
wrongly *accepts* is the failure that matters here, and it does not crash. It is
only the boundary that is misplaced.

I have not pushed anything to this branch. #101 is red for this and nothing
else; its other 24 checks are green.

Kartikey1306:
Fixed in `bbf4e89`, and the analysis above is confirmed in full — the oracle was mine and it was off by one. `eos_recovery_write_in_range()` is correct: its `len - 1u > UINT32_MAX - (base + offset)` deliberately bounds the **last written byte**, while the oracle asserted on the one-past-the-end address, so a write ending exactly on `0xFFFFFFFF` trapped despite being legal.

Verified three ways on the new head:

- **The crashing unit above, verbatim** (base=0xFFFFDBDB, slot=0x0001DBDB, offset=0, len=9253) → no trap, exit 0, under ASan+UBSan
- **A fail-open stub** accepting every input → trap still fires (exit 133), so the oracle kept its teeth; only the boundary moved
- 500,000 random inputs → clean

Worth recording why the job caught this when local testing did not: 500k *random* inputs never hit it, because random 32-bit values essentially never land `base+offset+len` exactly one past 2^32. Coverage-guided search walked to the boundary in six seconds. That is the argument for #101's job in one sentence — an oracle needs the same boundary analysis as the code it judges, and only execution finds the corner where it doesn't.

#101 rebasing onto `bbf4e89` should clear its last red check.

srpatcha (commented):
## Review — eBoot#99 "fix(fuzz): point every harness at the API it claims to fuzz"
head: bbf4e89c22d966bf8acc7f0579a7d2d16f8af709  author: Kartikey1306  ci: pass (24 checks, none of them a fuzz check)

**Verdict:** The diagnosis is right and I confirmed it mechanically rather than by reading. Against `libeboot_core.a` built from master, `nm -g --defined-only` reports `eos_bootctl_parse`, `eos_recovery_parse_packet`, `eos_fw_update_init` and `eos_fw_update_process_chunk` **absent** — those three harnesses could never have linked. `eos_image_parse_header` is **present**, and its real signature is `(uint32_t addr, eos_image_header_t *out)` (`include/eos_image.h:180`) against the harness's `(const void *, size_t)`, so that one linked and ran with a pointer where an address goes. The replacements are pointed at real entry points and the shared `fuzz_sim_flash.h` backend is the right call — I measured it: 200,000 inputs through the new `fuzz_bootctl` produce 200,000 flash reads, **none rejected**, all of them reading fuzzer bytes. Compare the #100 rewrite of the same file, which reads zero (see that review).

The corrected recovery oracle is right. `eos_recovery_write_in_range` (`core/recovery.c:260`–`273`) bounds the **last written byte** via `len - 1u > UINT32_MAX - (base + offset)`, and `recovery_handle_write` calls `eos_hal_flash_write(base + offset, buf, len)` — so `len - 1` is the correct displacement and the harness's `last > 0xFFFFFFFFULL` at `:50` now mirrors the implementation clause for clause.

### Findings
|---|---|---|---|---|
| 1 | Medium | — (CI) | **This head still builds none of the harnesses.** `EBLDR_BUILD_FUZZ` defaults `OFF` (`CMakeLists.txt:29`) and no workflow sets it — I grepped all 16 files under `.github/workflows` at `bbf4e89c` for `fuzz`/`FUZZ`: zero hits — and `checks.txt` lists 24 checks with no fuzz check. So the repair is correct and unenforced: the condition that hid four broken harnesses is untouched by this PR. The job is in **#101**, one commit on top of this one. | Land #99 and #101 together, or say in the body that #99 alone leaves the harnesses unbuilt. As it stands the PR title's promise ("point every harness at the API it claims to fuzz") is only compiler-checked once #101 lands. |
| 2 | Medium | — (duplicate) | **#100** is an independent single-commit implementation of the same repair by the same author, opened the same day, and it also adds the CI job. #99+#101 and #100 cannot both land. #99/#101 is the better of the two — it factors the flash backend into `tests/fuzz/fuzz_sim_flash.h` instead of copying it into each harness, and #100's own `fuzz_bootctl` does not reach the code it fuzzes. | Close #100 in favour of #99+#101, and say so on #100 so its green fuzz job is not read as a reason to prefer it. |
| 3 | Medium | `tests/fuzz/fuzz_sim_flash.h:78` | `memcpy(&fuzz_flash[addr], data, n)` is reached with `data == NULL`, from `fuzz_fw_update.c:39` (`fuzz_flash_load(0x4000, NULL, 0)`). `memcpy` declares both pointers non-null regardless of `n`, so this is undefined behaviour in a test helper. Reproduced: `gcc -fsanitize=undefined -fno-sanitize-recover=all`, one 8-byte input → `tests/fuzz/fuzz_sim_flash.h:78: runtime error: null pointer passed as argument 2, which is declared to never be null`, exit 1. It is invisible in the new job because `FUZZ_FLAGS` is `fuzzer,address` with no `undefined` (`tests/fuzz/CMakeLists.txt:17`) — which means the obvious next hardening step, adding `undefined`, would trip on every single input of that harness. | `if (n) memcpy(&fuzz_flash[addr], data, n);` in `fuzz_flash_load`, then add `undefined` to `FUZZ_FLAGS`. |
| 4 | Low | `tests/fuzz/fuzz_recovery_protocol.c:6`, header comment | The file is named for the recovery *protocol* and fuzzes one pure predicate inside it. `recovery_handle_write` (`core/recovery.c:275`) is the function that takes the wire fields, and `eos_recovery_enter` is the command loop; both remain unfuzzed, and `recovery_handle_write` is `static` so no harness can reach it. `.ai/security.md` puts OTA payloads and IPC frames among the parsers that "get fuzz coverage, not just unit tests". The other two rewritten harnesses state honestly what they drive; this one's name implies more than it does. | Say in the header comment that only the bounds predicate is covered and that the command loop is not. If protocol coverage is wanted, that needs a non-static seam, and it is a separate PR. |
| 5 | Low | `tests/fuzz/fuzz_crypto.c:16` | `/* Forward-declare crypto APIs */` is still there, immediately above the `#include "eos_crypto_boot.h"` that replaced the forward declarations it describes. | Delete the comment. |

The oracle in `fuzz_recovery_protocol.c` traps only on a write it wrongly **accepts**, never on one it wrongly **rejects**. That is deliberate and stated at `:38`–`:40`, and it is the right asymmetry for a bounds check — noting it so the one-sidedness is on the record, not as a finding.

### Architecture conformance
Conforms. Tier 1 Foundation (§21); everything is under `tests/`, no runtime dependency added, no include pointing up a tier (§5.1). `fuzz_sim_flash.h` installs board ops through `eos_hal_init` rather than defining `eos_hal_flash_read` itself, which keeps the harnesses on the same HAL seam the unit tests use — correct under `.ai/architect.md` ("`hal/` contains no board specifics"; a test fake belongs behind the interface, not as a duplicate symbol). §8 wants the eBoot TCB auditable, and a harness that names a function which does not exist is the opposite of auditable.

### Proposed changes
1. `if (n) memcpy(...)` in `fuzz_flash_load`; then add `undefined` to `FUZZ_FLAGS`.
2. Delete the stale comment in `fuzz_crypto.c`.
3. Bound the claim in `fuzz_recovery_protocol.c`'s header comment.
4. Resolve #99+#101 vs #100 before any of them lands.

### Not checked
- **libFuzzer link and run: NOT RUN here.** All six harnesses *compile* at this head; this host has no libFuzzer/ASan runtime (`libclang_rt.fuzzer.a`, `libclang_rt.asan.a` absent), so none links. The coverage numbers I quote come from a plain `gcc` driver of my own calling `LLVMFuzzerTestOneInput` in a loop over pseudo-random inputs, not from libFuzzer. That measures whether the harness reaches the code; it does not measure what coverage-guided search would find.
- **The 20.5M-execution run reported on #100, and #101's `Fuzz Harness Build`: not reproduced.** I read the check states via `gh pr checks`; I did not fetch or verify the job logs.
- **`fuzz_crypto`, `fuzz_image_verify`, `fuzz_fw_update` reach-through: NOT measured.** I instrumented `fuzz_bootctl` only. `fuzz_fw_update` returns early whenever `eos_fw_update_begin` fails and I did not check how often that is — worth measuring before trusting its execution count as coverage.
- **`ctest` on this head: NOT RUN.** The `core/ed25519_verify.c`, `include/eos_image.h` and `tests/unit/test_ed25519.c` parts of this diff are #94's commit (`f704d87f`), reviewed under #94/#98; I verified separately that master does not compile without them.
- The local clone does not have `bbf4e89c`; everything above is from a `gh api tarball` snapshot at that sha.


---
*Automated architecture review of `bbf4e89c22d9` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #107, opened 2026-09-04
fix(imgpack): reject version components that overflow the uint32 encoding

oscar-leung:
## Issue

`parse_version()` in `tools/imgpack.py` packed `major.minor.patch` into the header's `image_version` field without checking that each component fits the `EOS_VERSION_MAKE` layout (8/8/16 bits). Out-of-range input was silently misencoded: `1.256.0` became indistinguishable from `2.0.0`, `1.0.65536` from `1.0.0` (numerically older than `1.0.1`), and a negative patch like `1.0.-1` became patch 65535 — the maximum. Since the bootloader compares `image_version` numerically for anti-rollback (`eos_image_check_rollback`, `eos_image_check_version`) and raises a fuse-backed floor on confirm, a misencoded version can wrongly block legitimate updates or wrongly raise the floor above future builds.

## Approach

Validate each component against its field width in `parse_version()` and raise `ValueError`; `main()` reports it cleanly and exits 1 before any output is written. In-range versions encode byte-for-byte identically to before.

## Testing / validation

- Static review of the encoding contract against `include/eos_types.h` and the rollback call sites; both changed files pass `py_compile`.
- Added `tests/unit/test_imgpack.py` (pytest, subprocess-style matching the existing `test_sign_image.py`) covering correct encoding at the boundaries and rejection of the overflow/negative cases.
- The tests were written for CI to run; they were not executed locally.

## Limitations

Fixes only the tool-side encoding; the C `EOS_VERSION_MAKE` macro still masks rather than rejects (acceptable there — its inputs are compile-time constants). `eos_sign.py --version` takes a raw integer and is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/unit/test_imgpack.py
+++ b/tests/unit/test_imgpack.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 EoS Project
+
+"""Tests for imgpack.py version encoding.
+
+The bootloader compares image_version as a plain uint32 for anti-rollback
+(eos_image_check_version / eos_image_check_rollback), using the layout from
+EOS_VERSION_MAKE in include/eos_types.h: 8-bit major, 8-bit minor, 16-bit
+patch. A component that does not fit its field must be rejected: silently
+masking or carrying it produces a header that misorders against other images
+(e.g. '1.256.0' used to encode identically to '2.0.0', and '1.0.65536'
+identically to '1.0.0' — numerically *older* than 1.0.1).
+"""
+
+import struct
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+TOOLS = REPO_ROOT / "tools"
+
+IMAGE_VERSION_OFFSET = 20
+
+
+def _pack(tmp_path, version):
+    (tmp_path / "fw.bin").write_bytes(b"\x5A" * 256)
+    return subprocess.run(
+        [sys.executable, str(TOOLS / "imgpack.py"),
+         "--input", str(tmp_path / "fw.bin"),
+         "--output", str(tmp_path / "fw.eimg"),
+         "--load-addr", "0x08010000", "--entry-addr", "0x08010100",
+         "--version", version],
+        capture_output=True, text=True)
+
+
+def test_in_range_version_encodes_correctly(tmp_path):
+    r = _pack(tmp_path, "1.2.3")
+    assert r.returncode == 0, r.stderr
+    data = (tmp_path / "fw.eimg").read_bytes()
+    encoded = struct.unpack_from("<I", data, IMAGE_VERSION_OFFSET)[0]
+    assert encoded == (1 << 24) | (2 << 16) | 3
+
+
+def test_max_components_encode_correctly(tmp_path):
+    r = _pack(tmp_path, "255.255.65535")
+    assert r.returncode == 0, r.stderr
+    data = (tmp_path / "fw.eimg").read_bytes()
+    encoded = struct.unpack_from("<I", data, IMAGE_VERSION_OFFSET)[0]
+    assert encoded == 0xFFFFFFFF
+
+
+@pytest.mark.parametrize("version", [
+    "256.0.0",      # major does not fit 8 bits
+    "1.256.0",      # minor carries into major: used to equal 2.0.0
+    "1.0.65536",    # patch truncated to 0: used to encode older than 1.0.1
+    "-1.0.0",       # negative major
+    "1.0.-1",       # negative patch: used to encode as patch 65535
+])
+def test_out_of_range_version_is_rejected(tmp_path, version):
+    r = _pack(tmp_path, version)
+    assert r.returncode != 0, (
+        f"version '{version}' was accepted:\n{r.stdout}")
+    assert not (tmp_path / "fw.eimg").exists(), (
+        f"output image was written for rejected version '{version}'")
+    assert "out of range" in r.stderr
--- a/tools/imgpack.py
+++ b/tools/imgpack.py
@@ -31,12 +31,24 @@ SIG_CRC32 = 1
 
 
 def parse_version(version_str: str) -> int:
-    """Parse 'major.minor.patch' into uint32 encoding."""
+    """Parse 'major.minor.patch' into uint32 encoding.
+
+    Matches EOS_VERSION_MAKE in include/eos_types.h: 8-bit major, 8-bit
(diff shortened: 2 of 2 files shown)


srpatcha (commented):
## Review — eBoot#107 "fix(imgpack): reject version components that overflow the uint32 encoding"
head: 274d306c7433c7ec239953dedae4ffc3779a260b  author: oscar-leung  ci: pending (no check runs in bundle; `mergeStateStatus: BLOCKED`)

**Verdict:** The fix itself is correct and well-motivated — `image_version` is compared as a plain `uint32` by `core/image_verify.c:222` and `:230`, so a silently carried or truncated component really can misorder anti-rollback decisions — but one of the new parametrized test cases fails on every Python this repo's CI actually uses, so the PR as it stands turns the `test` job red.

### Findings
|---|---|---|---|---|
| 1 | High (P1) | `tests/unit/test_imgpack.py:65` (`"-1.0.0"` param) | `_pack` passes the version as two argv items (`"--version", "-1.0.0"`). On Python ≤ 3.13 `argparse._negative_number_matcher` is `^-\d+$\|^-\d*\.\d+$`, which `-1.0.0` does not match, so argparse classifies it as an unknown option and exits 2 with a usage message *before* `parse_version` runs. The third assertion, `assert "out of range" in r.stderr`, then fails. Reproduced under Python 3.12.14: `rc=2`, stderr is the argparse usage block. Python 3.14 loosened the matcher to `-\.?\d`, which is why it passes locally. `.github/workflows/ci.yml:26` runs the `test` job on `ubuntu-22.04` with no `setup-python`, i.e. system Python 3.10; `nightly.yml:83` pins 3.12. Both fail. Same defect makes the "negative major" case not exercise the validator at all even where it passes. | In `_pack`, join the flag and value: `f"--version={version}"`. Verified: that form gives `rc=1` and `out of range` in stderr on both 3.12.14 and 3.14.4. |
| 2 | Low (P3) | `include/eos_types.h:107` | `EOS_VERSION_MAKE` still masks (`(patch) & 0xFFFF`) and lets `major`/`minor` carry, i.e. the identical misencoding the PR is fixing on the tool side remains reachable in C. The PR body calls this out and argues the inputs are compile-time constants; that is true today — the only in-tree use is `tests/unit/test_slot_manager.c:214` — but nothing enforces it. | Optional follow-up, not for this PR: make the macro reject at compile time, e.g. wrap the operands in a `static_assert`-backed helper, or add a `_Static_assert` in `core/` for each literal call site. |
| 3 | Low (P3) | `tools/imgpack.py:87`, `docs/quickstart.md:90` | The accepted range is now enforced but not documented anywhere a user looks before they hit the error: `--version` help is still `'Image version (major.minor.patch)'`, and the quickstart example does not mention field widths. Behaviour changed without the surrounding docs changing with it. | Extend the argparse help to `'Image version major.minor.patch (major/minor 0-255, patch 0-65535)'` and add one sentence to `docs/quickstart.md`. |
| 4 | Low (P3) | `tests/unit/test_imgpack.py:31`, `tests/unit/test_sign_image.py:34,45` | `IMAGE_VERSION_OFFSET = 20` and the `subprocess.run([sys.executable, ...], capture_output=True)` runner are now defined twice in `tests/unit/`. Both are transcriptions of `include/eos_image.h`; the ABI test (`tests/unit/test_image_abi.c:69`) is a third copy. | Move the offsets and the runner into a `tests/unit/conftest.py` and import from both files. |

### Architecture conformance
Conforms. Per master design §5.1 and §8, `tools/imgpack.py` is a host-side packaging tool, not part of the eBoot trusted computing base — the change adds no code to `stage0/`, `stage1/` or `core/`, so the TCB surface is unchanged and the "minimal and auditable" constraint is untouched. The change strengthens §8.1's "rollback protection where hardware/policy supports it" by making the producer refuse to emit a header the consumer's ordering comparison would misread; it does not alter that comparison. No dependency direction is affected: no new import, link, or manifest entry.

§23.2 (firmware-format compatibility) is satisfied and I verified it rather than assuming it: for in-range inputs the old and new tools emit byte-identical images (SHA-256 of the full output matches for `0.0.0`, `1.2.3` and `255.255.65535`). The only behavioural change is that previously-accepted out-of-range input is now an error with exit 1 — the PR body states this explicitly, so the compatibility impact is not silent. §28: the PR body's evidence claims are accurate and it says plainly that the tests were not executed locally, which is exactly why finding 1 was not caught.

Tier placement (§21): eBoot is Tier 1 - Foundation, and a packer for the eBoot image format belongs in eBoot. Correct repo.

### Proposed changes
One edit, in `tests/unit/test_imgpack.py`:

```python
     return subprocess.run(
         [sys.executable, str(TOOLS / "imgpack.py"),
          "--input", str(tmp_path / "fw.bin"),
          "--output", str(tmp_path / "fw.eimg"),
          "--load-addr", "0x08010000", "--entry-addr", "0x08010100",
-         "--version", version],
+         f"--version={version}"],
         capture_output=True, text=True)
```

`--opt=value` bypasses the option-vs-negative-number heuristic on every supported Python, so both `-1.0.0` and `1.0.-1` reach `parse_version`. Worth a one-line comment on that line saying why, or the next person will "tidy" it back.

Findings 2–4 are follow-ups; none should hold this PR.

### Not checked
- **The PR's own pytest suite was NOT RUN as written.** `pytest` is not installed on this host and `python3 -m venv` fails here (no `ensurepip`). I exercised `tools/imgpack.py` directly with the patch applied in a scratch tree (`git apply` clean), driving every one of the 7 parametrized/explicit cases through `subprocess` the same way the test does, under both Python 3.14.4 and 3.12.14. That is what finding 1 rests on; I did not observe pytest's own report.
- I did not confirm the exact Python minor version on the `ubuntu-22.04` GitHub runner. `ci.yml` installs no `setup-python`, so it is the image's system `python3`; the argparse matcher was loosened in 3.14, and 3.12.14 already fails, so any runner Python below 3.14 fails. Inferred, not observed on the runner.
- **CI result: unknown.** `checks.txt` in the bundle is empty — the PR was opened one minute before the run started, so no check runs had been reported. `mergeStateStatus` is `BLOCKED`, which is consistent with required checks not yet complete; I did not verify the branch-protection rules.
- No C build and no `ctest` run. The diff touches no C, and the C-side encoding path was read, not executed.
- I did not review `tools/eos_sign.py`'s `--version` handling beyond confirming the PR's claim that it takes a raw integer.


---
*Automated architecture review of `274d306c7433` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #110, opened 2026-09-10
fix(stage0): return after failed stage1 hash verification

Vivekmittal0404:
## Summary

Fix fail-open in Stage-1 integrity check — a hash mismatch was logged and routed to recovery, but execution still fell through and jumped into the unverified Stage-1 image if eos_recovery_enter() returned.

## Type of Change
 — Bug fix

## Changes
stage0/jump_stage1.c: add return; after eos_recovery_enter() in the mismatch branch, so a failed integrity check can no longer fall through to ops->jump(stage1_addr).

## Testing
Manual testing performed

## Related Issues
 Related to the fail-open fixes in secure_boot.c and jump_app.c

### DIFF
--- a/stage0/jump_stage1.c
+++ b/stage0/jump_stage1.c
@@ -94,6 +94,11 @@ void ebldr_stage0_main(void)
         if (match1 || match2) {
             eos_boot_log_append(EOS_LOG_BOOT_FAIL, EOS_SLOT_NONE, 0xBAD1);
             eos_recovery_enter(&bctl);
+            /* eos_recovery_enter() only returns if it decides to reboot rather
+            * than halt. Either way, a Stage-1 that just failed its integrity
+            * check must never be jumped into — return so the corrupted image
+            * is never entered. */
+           return;
         }
         eos_boot_log_append(EOS_LOG_IMAGE_VALID, EOS_SLOT_NONE, 0);
 #endif


srpatcha (commented):
## Review — eBoot#110 "fix(stage0): return after failed stage1 hash verification"
head: f163b063504fa63a519a6c1b2a9b1afdd792aed1  author: Vivekmittal0404  ci: none reported

**Verdict:** The one-line `return` is the correct fail-closed behaviour and should land — but the comment and PR body justify it with a claim about `eos_recovery_enter()` that the code contradicts, and nothing about this change has been compiled or tested by anything.

### Findings
|---|---|---|---|---|
| 1 | Medium | `stage0/jump_stage1.c:96-99` | The added comment asserts "`eos_recovery_enter()` only returns if it decides to reboot rather than halt". It never returns. `core/recovery.c:461` is a bare `while (1)` with no `break` and no `return` inside it; every command that leaves recovery does so through `eos_hal_system_reset()` (`core/recovery.c:351`, `:506`), and the `return EOS_OK;` at `core/recovery.c:523` is unreachable. So the fall-through the PR body describes — "execution still fell through and jumped into the unverified Stage-1 image if `eos_recovery_enter()` returned" — cannot occur in the current tree. **Keep the change**: it is correct defense-in-depth and it is what makes the fail-closed property hold independently of the callee. But this is TCB code, and a comment stating a contract the callee does not have will be relied on by the next reader. | Reword to state the real situation, e.g. "`eos_recovery_enter()` does not return today (`core/recovery.c:461`); returning here makes the fail-closed behaviour independent of that." Then give the callee the contract explicitly — declare `eos_recovery_enter()` `__attribute__((noreturn))`, or document why it keeps an `int` return, and delete the unreachable `return EOS_OK;`. The stale comment at `stage0/jump_stage1.c:54` says the same wrong thing and should be fixed in the same pass. |
| 2 | Medium | PR CI | No checks ran at all. `gh pr checks 110` reports "no checks reported on the 'fix/stage1-verify-fallthrough' branch", the bundle's `checks.txt` is empty, and the PR is `BLOCKED`. Nothing here has been compiled. Note that eBoot `master` cannot configure at all right now — `tests/CMakeLists.txt` defines `eboot_test_fdt_loader` / `test_fdt_loader` twice (lines 187-191 and 201-205) — so even a re-run would go red for an unrelated reason until that is removed (see the review on #109). | Push an empty commit or re-trigger once the duplicate CMake target is removed from `master`, and post the resulting cross-compile output. |
| 3 | Medium | PR body, "Testing" | "Manual testing performed" — no board named, no command, no output, no result. `.ai/reviewer.md`: every PASS should point at command output, and an unsupported claim is itself the finding. `.ai/security.md` holds boot-chain code to a stricter standard than the rest of the tree, so this is the wrong file to assert an unevidenced pass on. | Either state what was actually done (board, `EBLDR_BOARD=`, how the mismatch was induced, what was observed) or say plainly that it was not tested and why. |
| 4 | Medium | `CMakeLists.txt:360-366` | There is no regression test and, as the tree stands, there cannot be one: `stage0/jump_stage1.c` is compiled only inside the `ebldr_stage0` target, which is added only in the board cross-compile branch. The host build is "Core libraries only — eboot_hal, eboot_core, eboot_stage1", so no host `ctest` binary links `ebldr_stage0_main()` and nothing will catch a reintroduction of this fall-through. The repo already has the pattern that would cover it — `tests/unit/test_jump_app_bounds.c` drives a simulated flash and asserts the jump never happens — but it can only reach `stage1/jump_app.c`, not stage 0. | Out of scope for a one-line fix, but worth an issue: make the stage-0 logic host-testable (compile `stage0/jump_stage1.c` into a host-only object with the board `ops` mocked, as `test_jump_app_bounds.c` already does for stage 1), then add a test that a hash mismatch never reaches `ops->jump`. |
| 5 | Low | `stage0/jump_stage1.c:96-100` | Indentation does not match the surrounding block: `return;` is indented 11 spaces inside a 12-space block, and the comment continuation lines put ` *` one column left of the opening `/*`. | Reindent to 12 spaces and align the continuation asterisks. |

### Architecture conformance
Conforms. Master design §5.1 — "eBoot keeps the trusted computing base minimal and auditable": the change adds one statement to stage 0 and no dependency of any kind, so the TCB surface is unchanged. §8's boot sequence is unaffected in ordering; the change only removes a path by which a Stage-1 image that failed `Verify Image` could still be entered. Per `.ai/architect.md`'s target shape, `stage0/` is the correct home for "minimum hardware init, smallest possible surface" and this belongs there rather than in `core/`. No tier question arises (§21: eBoot is Tier 1). No public interface, struct layout, manifest field or serialized format changes, so brief §8 does not apply.

### Proposed changes
```diff
--- a/stage0/jump_stage1.c
@@
         if (match1 || match2) {
             eos_boot_log_append(EOS_LOG_BOOT_FAIL, EOS_SLOT_NONE, 0xBAD1);
             eos_recovery_enter(&bctl);
-            /* eos_recovery_enter() only returns if it decides to reboot rather
-            * than halt. Either way, a Stage-1 that just failed its integrity
-            * check must never be jumped into — return so the corrupted image
-            * is never entered. */
-           return;
+            /* eos_recovery_enter() does not return today (core/recovery.c:461
+             * loops forever; every exit resets the SoC). Returning here makes
+             * the fail-closed behaviour independent of that: a Stage-1 that
+             * failed its integrity check must never be jumped into. Control
+             * unwinds to Reset_Handler's while(1) and the watchdog resets. */
+            return;
         }
```
Then, in the same PR or a follow-up: add `__attribute__((noreturn))` to the `eos_recovery_enter()` declaration at `stage0/jump_stage1.c:22` and its definition, drop the unreachable `return EOS_OK;` at `core/recovery.c:523`, and correct the comment at `stage0/jump_stage1.c:54`. Nothing else needs to move.

### Not checked
- **Nothing was built or run.** No CI ran on this branch, and eBoot `master` does not currently configure, so no build evidence exists from any source. Every finding above is from reading the head revision `f163b06`, not from observed output.
- That control actually unwinds to `Reset_Handler`'s `while (1)` and that the watchdog then resets: read at `stage0/reset_entry.c:77-80`, but never executed on hardware or in simulation by this review.
- Whether any board port overrides or wraps `eos_recovery_enter()` such that it *can* return — only the definition in `core/recovery.c` was examined; `boards/` was not swept.
- Behaviour with `EBLDR_VERIFY_STAGE1=OFF`, where this entire block compiles out. Not exercised.


---
*Automated architecture review of `f163b063504f` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (approved):
Thanks Vivek — agree with the change. `eos_recovery_enter()` is a `while (1)` today so the fall-through is not currently reachable, but its `int` return and the "does not return unless recovery instructs a reboot" comment at `stage0/jump_stage1.c:55` both say it may, and a Stage-1 that just failed its hash must never be reached by any path. The explicit `return` makes that fail-closed independent of recovery's future behaviour, and `reset_entry.c:80` parks in `while (1)` afterwards.

One cosmetic ask at `stage0/jump_stage1.c:97-101`: the comment continuation lines and the `return;` are misaligned (11 spaces vs the block's 12). Please match the file's 4-space style.

Approving with that nit.

PR #123, opened 2026-09-14
fix(sha512): drop a carry term that is always zero

### LINKED ISSUE #121: sha512.c: the `copy >> 61` carry term in eos_sha512_update() is always zero
## Symptom
`core/sha512.c`, `eos_sha512_update()`: `ctx->bitlen[0] += (uint64_t)copy >> 61;` is always zero. `copy` is clamped to `128 - ctx->buffer_len`, so `copy <= 128` and `copy >> 61 == 0` on every iteration. The real 128-bit carry is already handled by the `old_low` comparison two lines above.

Not a defect — the digest is correct (RFC 8032 vectors and the SHA-512 KATs pass) — but the line reads as overflow handling and does nothing, which is the kind of thing a future reader preserves because they assume it matters.

Raised in review of #115; left out of that PR deliberately so the file stayed a byte-exact restoration of the previously reviewed blob (`d9aa57c`).

Kartikey1306:
**Stacked on #115.** Review `e152d8e..46e1d89` (corrected from `a276016..b828d02` at 12:20 IST 2026-09-15: #115 was rebased onto master `682d005` as `e152d8e` and this branch restacked onto it — one commit replayed clean, content unchanged; verified on the new head before the push: build OK, ctest 31/31, pytest 58 passed / 3 skipped): one commit, one file, 2 insertions / 1 deletion against blob `d9aa57c`.

## Problem (review finding 2 on #115; eBoot #121)

`core/sha512.c`, `eos_sha512_update()`:

```c
ctx->bitlen[0] += (uint64_t)copy >> 61;
```

is always zero. `copy` is clamped to `128 - ctx->buffer_len`, so `copy <= 128` and `copy >> 61 == 0` on every iteration. The real 128-bit carry is already handled by the `old_low` comparison two lines above. The digest is correct either way (RFC 8032 vectors and the SHA-512 KATs pass), but the line reads as overflow handling and does nothing — the kind of thing a future reader preserves because they assume it matters.

Kept out of #115 deliberately so that `core/sha512.c` there stays a byte-exact restoration of the previously reviewed blob.

## Change

The line is replaced by a two-line comment giving the bound: `copy` is at most `128 - buffer_len`, so `copy << 3` fits the low word and can wrap it at most once per iteration; the `old_low` comparison above is that carry.

## Verification

**Verified**, macOS/clang: Release build clean, `ctest` 31/31, `eboot_test_ed25519` 14/14 including `test_sha512_known_answers` and `test_sha512_streaming_matches_one_shot` (the streaming test drives `eos_sha512_update()` across chunk boundaries, which is where a wrong carry would show).

## Closing issue

Fixes #121


srpatcha (commented):
## Review — eBoot#123 "fix(sha512): drop a carry term that is always zero"
head: b828d02feb9b1a5c18056e2c1a5d0dab94d03113  author: Kartikey1306  ci: pass (26 green, 0 red; only `Create GitHub Release` skipped)

**Verdict:** Clean. One dead line removed from `core/sha512.c` and replaced with a comment
that states the invariant making it dead. The reasoning is correct, and I confirmed the change
is a behavioural no-op by hashing 360 inputs through the code before and after and diffing the
output: byte-identical, and both match the reference implementation. Nothing to fix.

### Findings
|---|---|---|---|---|
| — | — | — | None. | — |

This PR is one deleted line and one added comment. I verified it rather than waved it through
— evidence below — and there is nothing to report. Per the brief, that is the whole finding
table; I am not manufacturing a second opinion to fill it.

**Disclosure, because it matters for who is checking whom:** this deletion was recommended by
this pipeline's own earlier review of #115 (finding 2). `.ai/reviewer.md`'s rule is that you do
not approve your own work; suggesting a change in a review is not implementing it, but it does
mean I had a prior answer in mind, so I verified this one harder than its size warrants rather
than confirming myself. The evidence below stands on its own and would have caught a wrong
deletion.

### Why the line was dead
`core/sha512.c:190` sets `copy = 128 - ctx->buffer_len`, then clamps it down to `len`.
`buffer_len` is `0` after `eos_sha512_init()` and is reset to `0` immediately after every
`sha512_transform()` (`:214`), so it is in `[0,127]` on entry to the loop body and `copy` is in
`[1,128]`. `128 >> 61 == 0`, so `(uint64_t)copy >> 61` was zero on every iteration the function
can ever execute — not "in practice", but for every reachable value. The real 128-bit carry is
the `old_low` comparison at `:206-207`: `bits = copy << 3 <= 1024`, so adding it to
`bitlen[1]` can wrap the low word at most once, and that single wrap is exactly what the
comparison detects and propagates into `bitlen[0]`. The replacement comment says this, which is
the useful outcome — the line's problem was that it *looked* like overflow handling, so a
future reader would preserve it on the assumption it mattered.

There is no test to add here and correctly none was added: a test that distinguishes the two
versions would have to hash more than 2⁶¹ bytes.

Detached worktree at `.ai/autoreview/state/scratch/eBoot-123` off `refs/pull/123/head`. The

| Check | Result |
|---|---|
| `ctest --test-dir build/host --output-on-failure --no-tests=error` | **PASS — 31/31**, 10.92s |
| `eboot_test_ed25519` verbose | **PASS — 14/14**, including `test_sha512_known_answers` and `test_sha512_streaming_matches_one_shot`. Ed25519 derives its challenge scalar from this hash, so the RFC 8032 vectors would fail on any deviation. |
| **Differential test against a reference implementation.** Built a harness linking `libeboot_core.a` and hashed 20 lengths chosen to straddle every buffer and block boundary — `0,1,55,56,63,64,111,112,113,127,128,129,200,255,256,257,1023,1024,1025,4096` — one-shot, and again streamed in 17 different chunk sizes (1…17) each | **PASS — 360 digests, 0 mismatches.** Every one-shot digest equals Python `hashlib.sha512` of the same input, and every chunked digest equals its own one-shot. |
| **Is the deletion a no-op?** Restored `core/sha512.c` from the parent commit `a276016`, rebuilt `eboot_core`, re-ran the same harness, diffed the two outputs | **IDENTICAL** — the pre-fix and post-fix trees produce byte-identical digests across all 20 lengths × 18 chunkings. This is the claim that matters and it is now measured, not argued. `core/sha512.c` was restored afterwards; `git diff --stat` clean. |
| Can `copy` be `0`, making the loop spin? | **No** — `buffer_len` is reset to `0` on reaching 128 (`:215`), so it is never 128 on entry and `copy >= 1`. Checked because the deletion drew my eye to the loop bounds. |

**Was anything weakened?** No. Nothing else in the diff: no test disabled, no assertion
removed, no lint loosened. The change is `-1/+2` lines in one file, and the two added lines are
a comment.

### Architecture conformance
Conforms. §21: eBoot is Tier 1 Foundation and `core/sha512.c` is in the owning repo, in the
directory `.ai/architect.md`'s target shape assigns to shared boot logic. No `#include`, link
line, CMake entry or manifest dependency changes, so §5.1's dependency direction is untouched
and nothing points up a tier. §14.1's "do not invent cryptographic primitives" is not engaged —
this removes a line, it does not alter the algorithm, and the differential test above
demonstrates that directly. §5.1's "eBoot keeps the trusted computing base minimal and
auditable" is served in the small: dead code that impersonates a security-relevant operation is
the opposite of auditable.

**On keeping it out of #115.** The right call, and worth recording. #115 restores
`core/sha512.c` to blob `d9aa57c` byte-for-byte, and that byte-identity is itself the evidence
for #115's central claim — that the file was not hand-written into the TCB but restored from a
previously-reviewed blob. Making this edit inside #115 would have destroyed the one property
that made #115 cheap to trust. Splitting it into a one-line PR stacked on top preserves both.

### Proposed changes
```
None. Merge after #115 (it is stacked on a276016 and touches a file #115 restores).
```

### Not checked
  NOT RUN locally.** CI reports all green on this head; I did not read the logs. My
  differential harness is a host x86-64 build with gcc, so the result is **not** evidence about
  a big-endian or 32-bit `size_t` target — though on such a target `copy` is smaller still, so
  the deleted term is if anything more obviously zero.
- **`pytest tests/` — NOT RUN at this head.** The diff touches no Python. I ran it at #115's
  head (80 passed) and expect it unchanged; I did not execute it here.
- **ASan/UBSan and Valgrind — NOT RUN.** `test_ed25519` is in `EBLDR_UNIT_TESTS` so CI's
  Valgrind leg covers it there; I did not run Valgrind locally.
- **No behaviour at lengths above 2³² bytes was exercised.** The 128-bit length accumulator's
  high word `bitlen[0]` is only reached after 2⁶¹ bytes, which is not testable; the argument
  that the carry is handled correctly rests on reading `:205-207`, not on execution. That was
  equally true before this PR.


---
*Automated architecture review of `b828d02feb9b` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
Acknowledged — nothing to change. Head stays `b828d02`. Merge after #115, as stated: stacked on `a276016` and it touches `core/sha512.c`, which #115 restores.

srpatcha (commented):
## Review — eBoot#123 "fix(sha512): drop a carry term that is always zero"
head: 46e1d8910bca61d2dff0a88ac27ab82d6ffd9e92  author: Kartikey1306  ci: pass (26 green, 0 red; only `Create GitHub Release` skipped)

**Verdict:** The previous review of `b828d02f` found nothing, and there is still nothing. This
head is a **pure re-stack** onto #115's rebased tip — the single commit is byte-identical by
`git patch-id` — so no new work has been introduced to review. Re-verified green on the new
base rather than assumed. Merge after #115.

### Status of the previous review (`eBoot-123-b828d02f.md`)

No findings were raised, so there is nothing to follow up. The differential evidence that
carried it — 360 digests identical before and after the deletion, and identical to
`hashlib.sha512` — was gathered against parent `a276016`; `core/sha512.c` is unchanged between
that parent and this one, so the result carries. That is Inferred from `patch-id` rather than
re-measured; see "Not checked".

### Findings

|---|---|---|---|---|
| — | — | — | None. | — |

One re-stacked commit, `-1/+2` lines in one file, two of which are a comment. Per the brief I
am not manufacturing a finding to justify commenting again.

Detached worktree at `.ai/autoreview/state/scratch/eBoot-123` off `refs/pull/123/head`. The
user's `eBoot` checkout was not touched; nothing was committed or pushed.

| Check | Result |
|---|---|
| Is this head new work or a re-stack? | **RE-STACK, confirmed.** One own commit at each head; `git patch-id --stable` of `b828d02f` and `46e1d891` are **identical**. The parent moved `a276016` → `e152d8e`, which is #115's own rebase. |
| `ctest --test-dir build/host --output-on-failure --no-tests=error` | **PASS — 31/31**, 11.41s |
| `eboot_test_ed25519` verbose | **PASS — 14/14**, including `test_sha512_known_answers` and `test_sha512_streaming_matches_one_shot`. Ed25519 derives its challenge scalar from this hash, so the RFC 8032 vectors would fail on any deviation in it. |
| Did the new base change `core/sha512.c` under the PR? | **No** — the commit still applies to blob `d9aa57c`, which is what #115 restores; the diff header reads `index d9aa57c..89bdb99`. |

**Was anything weakened?** No. There is no new source since the last review.

**One stale detail, not a finding.** The PR comment of 2026-09-15T06:19Z says "Head stays
`b828d02`. … stacked on `a276016`". Both are now out of date — the branch was re-stacked to
`46e1d891` on `e152d8e` when #115 rebased. The content is unchanged, so nothing about the
comment's substance is wrong; noted only so the next reader does not try to resolve a SHA that
is no longer reachable. Ranges on a three-deep stack go stale on every re-stack and chasing
them is not worth anyone's time.

### Architecture conformance

Conforms; re-checked, not recalled. §21: eBoot is Tier 1 Foundation and `core/sha512.c` sits in
the directory `.ai/architect.md`'s target shape assigns to shared boot logic. No `#include`,
link line, CMake `target_link_libraries` entry or manifest dependency changes, so §5.1's
dependency direction is untouched and nothing points up a tier. §14.1's "do not invent
cryptographic primitives" is not engaged: this removes a line and adds a comment, and the
previous review's differential test demonstrated the algorithm is unchanged. §5.1's "eBoot keeps
the trusted computing base minimal and auditable" is served in the small — dead code that
impersonates a security-relevant operation is the opposite of auditable.

### Proposed changes

```
None. Merge after #115.
```

### Not checked

- **The differential harness was NOT re-run at this head.** 360 digests × before/after was run at
  `b828d02f`; the commit is patch-identical here and applies to the same blob, so the result
  carries — but that is Inferred from `patch-id`, not re-observed. The passing
  `test_sha512_known_answers` and `test_sha512_streaming_matches_one_shot` at this head are the
  direct evidence I do have.
  NOT RUN locally.** CI green on this head; job logs not read. The local build is host x86-64
  gcc, so nothing here is evidence about a big-endian or 32-bit `size_t` target.
- **`pytest tests/` — NOT RUN at this head.** The diff touches no Python.
- **No behaviour above 2³² bytes was exercised.** `bitlen[0]` is only reached after 2⁶¹ bytes,
  which is not testable; the carry argument rests on reading `core/sha512.c:205-208`. That was
  equally true before this PR.


---
*Automated architecture review of `46e1d8910bca` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
Ack for the 06:49Z re-review at `46e1d89`: pure re-stack onto #115's `e152d8e`, single commit byte-identical, nothing open. Merge after #115.

PR #133, opened 2026-09-14
build: drop EBLDR_REQUIRE_SIGNATURES and EBLDR_RECOVERY_AUTH, switches wired to nothing

### LINKED ISSUE #132: EBLDR_REQUIRE_SIGNATURES and EBLDR_RECOVERY_AUTH are build options that no source file reads
## Where
`CMakeLists.txt:24-25` (`option(...)`), `CMakeLists.txt:57-62` (`add_compile_definitions(...)`), `README.md` "Security options" table, `docs/book/book.md` §24.1.

## What
Both options are offered with default `ON`, each is forwarded to the compiler as a definition, and the README presents them as the switches that require Ed25519 signatures and recovery authentication. Nothing reads them:

```
$ git grep -n "EBLDR_REQUIRE_SIGNATURES\|EBLDR_RECOVERY_AUTH" -- '*.c' '*.h'
(no output)
```

Measured: configuring the same tree with `-DCMAKE_C_FLAGS="-DEBLDR_REQUIRE_SIGNATURES -DEBLDR_RECOVERY_AUTH"` and without produces byte-identical output for all 41 non-test object files (`cmp` over `CMakeFiles/**/*.c.o`; only the archive timestamps differ). So `OFF` builds the verifying bootloader too -- fail-closed by accident, but:

- an integrator who reads `CMakeCache.txt` concludes that these two flags are what enforce verification; they are not, and a future toggle added "to match the option" would be a way to build a non-verifying bootloader;
- a bring-up script that sets either `OFF` gets no signal that the request was ignored;
- `tests/simulate_tests.py:172-173` "verify" the defaults by checking that the option names appear in `CMakeLists.txt`, which holds while the switch does nothing (that script is run by no workflow and already fails 9 unrelated checks on master).

`docs/book/book.md` §24.1 goes further and lists `EBLDR_SECURE_BOOT`, `EBLDR_MULTICORE`, `EBLDR_RECOVERY` and `EBLDR_BOOT_MENU` as build options. None has ever existed in `CMakeLists.txt`.

## Expected
Signature verification and recovery authentication are unconditional, and the build system and docs should say exactly that: no option that claims to disable them, a configure that refuses `OFF` instead of silently building the verifying firmware, doc tables that list only options that exist, and a guard so that a forwarded `EBLDR_` definition nothing reads cannot come back.

Kartikey1306:
Fixes #132

**Stacked on #135 (→ #115); review `abf0522..f9fa26c` for this change alone** -- two commits, six files. Merge order **#115 → #135 → this**: #135 is the Eisvogel toolchain fix for `Build PDF`, which this PR triggers by touching `docs/book/` and which has been red on every run since 2026-04-25; stacking on it is what turns that check green here rather than leaving it as a declared-prerequisite red. (Restacked from `e152d8e..25794f4` once #135's fix was proven green at `abf0522`; before that from `a276016..88cb92e`.) Merge #115 first. Independent of the other stacks on #115 (#116/#122, #123, #127/#131, #129): its CHANGELOG line goes at the end of `### Fixed`, where none of them writes.

**On the red `Build & Release Book PDF` check:** this PR is the first one tonight to touch `docs/book/`, so it is the first to trigger that workflow, and the workflow has never succeeded in this repository -- every run since 2026-04-25 fails, including the last `master` push run (`84bbef0`, 2026-09-08), on the same `! LaTeX Error: Command \chead already defined.` (pandoc 3.6.1 + Eisvogel 2.4.0). Not caused by the table edit here; the toolchain fix is #135 (issue #134), and this PR's PDF check goes green once that lands. The gate checks (`CI — eBoot`, `eBoot Build & Test`, `CodeQL`, `Linked issue policy`, `Simulation Sanity Test`) are green on this head.

## The defect
`CMakeLists.txt` offered `EBLDR_REQUIRE_SIGNATURES` and `EBLDR_RECOVERY_AUTH` as options (default `ON`), forwarded each as a compile definition, and the README listed both under "Security options" as the switches that require Ed25519 signatures and recovery authentication. No source file tests either macro:

```
$ git grep -n "EBLDR_REQUIRE_SIGNATURES\|EBLDR_RECOVERY_AUTH" -- '*.c' '*.h'
(no output)
```

Measured rather than inferred: the same tree configured with `-DCMAKE_C_FLAGS="-DEBLDR_REQUIRE_SIGNATURES -DEBLDR_RECOVERY_AUTH"` and without produces byte-identical output for **41 of 41** non-test object files (**to reproduce, build without `-g`** — debug info embeds the command line, so a Debug configuration differs in every object for a reason that has nothing to do with the options) (`cmp` over `CMakeFiles/**/*.c.o`; the `.a` archives differ only in their timestamp fields). `OFF` built the verifying bootloader too -- fail-closed by accident, but a documented switch that changes nothing is a capability the tree does not have, presented as configuration. `docs/book/book.md` §24.1 went further and listed `EBLDR_SECURE_BOOT`, `EBLDR_MULTICORE`, `EBLDR_RECOVERY` and `EBLDR_BOOT_MENU`, none of which has ever existed in `CMakeLists.txt`.

## The fix
- `CMakeLists.txt`: the two `option()`s and their `add_compile_definitions` are removed. A configure that still passes either as `OFF` stops with `FATAL_ERROR` ("there is no build that skips signature verification or recovery authentication ... drop the flag") instead of silently producing the verifying firmware. Passing `ON` is inert and gets CMake's usual unused-variable warning.
- `README.md`: the two rows are gone; a paragraph states that verification and recovery authentication are not options, and why the rows were there.
- `docs/book/book.md`: the table lists the options that exist (`EBLDR_BOARD`, `EBLDR_BUILD_TESTS`, `EBLDR_VERIFY_STAGE1`, `EBLDR_HARDENING`, `EBLDR_SANITIZE`, `EBLDR_BUILD_FUZZ`) and names the four phantoms it used to carry.
- `tests/unit/test_build_options_are_read.py` (new, runs under the existing `pytest tests/` step in `ci.yml`): every `EBLDR_` name that `CMakeLists.txt` forwards with `add_compile_definitions` must be referenced by a preprocessor line in some `.c`/`.h` under `core hal include stage0 stage1 boards tests`; every `EBLDR_` name in a README or book option table must be declared by `CMakeLists.txt`; and neither removed option may be redeclared.
- `tests/simulate_tests.py`: **a correctness repair to a file nothing executes — not coverage.** Its two checks asserted the option *names* appeared in `CMakeLists.txt`, which held *because* the switches did nothing, so each asserted the opposite of what its name claimed. They now assert the honest opposite, and they are exactly as unexecuted as before: no workflow, CMake target or Makefile runs this file (`grep -rn simulate_tests .github/ CMakeLists.txt tests/CMakeLists.txt` is empty), and run by hand it fails 8 unrelated stale checks. The guard that pytest actually runs is `tests/unit/test_build_options_are_read.py`, which the CHANGELOG names. The file's fate — register-and-repair, or delete — is a follow-up, not this PR. That script is run by no workflow and fails 9 unrelated checks on the base (`3.0.1` version pins, CI job names); the failure set is unchanged, 9 before and 9 after.

Not done: `EBLDR_VERIFY_STAGE1` is a real switch (`stage0/jump_stage1.c:65`) and stays one; whether a Release build of a real board should be allowed to turn it off is a policy question for #122's gate, not this PR.

## Verification
| Check | Result |
|---|---|
| `cmake -S . -B build -DEBLDR_BUILD_TESTS=ON` → `cmake --build` → `ctest` | **31/31** |
| `cmake ... -DEBLDR_REQUIRE_SIGNATURES=OFF` / `-DEBLDR_RECOVERY_AUTH=0` | configure **fails** at the new guard with the message above |
| `cmake ... -DEBLDR_REQUIRE_SIGNATURES=ON` | configures (exit 0) |
| `pytest tests -q` | **84 passed** (80 on the base + the 4 new) |
| **Negative control:** base `CMakeLists.txt`, `README.md` and `docs/book/book.md` restored, new guard run | **3 of 4 fail**: unread definitions `['EBLDR_RECOVERY_AUTH', 'EBLDR_REQUIRE_SIGNATURES']`; phantom book options `['EBLDR_BOOT_MENU', 'EBLDR_MULTICORE', 'EBLDR_RECOVERY', 'EBLDR_SECURE_BOOT']`; the removed options still declared |
| Object comparison, definitions forced back in vs. not | 41/41 identical |
| Line endings | `CMakeLists.txt` stays CRLF (464→471 CR bytes), `CHANGELOG.md` 135→136; README, book, and the Python files stay LF |


srpatcha (commented):
## Review — eBoot#133 "build: drop EBLDR_REQUIRE_SIGNATURES and EBLDR_RECOVERY_AUTH, switches wired to nothing"
head: 88cb92e9b262da1d7c71f1a5531dcc1102d267d3  author: Kartikey1306  ci: **fail** — `Build PDF` red, 25 green (`Create GitHub Release`, `Attach PDF to Release`, `Update Dev PDF`, `assign` skipped)

**Verdict:** First look. The claim this PR rests on is true and I reproduced it exactly: with both
macros forced back in via `CMAKE_C_FLAGS`, **41 of 41 objects are byte-identical** to a build without
them, so the two options documented a capability the tree did not have. Removing them rather than
implementing them is the right call — the alternative is adding a supported way to build a
bootloader that does not verify. The `FATAL_ERROR` for a script that still passes `=OFF` is the
right shape, and it fires. Two findings, neither in the removal itself: the new guard catches one of
the three ordinary CMake spellings of the mistake it exists to prevent, and `tests/simulate_tests.py`
— which this PR carefully corrects — is executed by nothing.

**The red check is not this PR's.** `Build PDF` fails with `LaTeX Error: Command \chead already
defined` at `l.1560` of the template preamble, which is a pandoc/Eisvogel/`fancyhdr` incompatibility
that never reaches the document body. **#135** in this same batch is the PR that fixes it. I read the
job log rather than inferring this, because this PR does edit `docs/book/book.md` and that was the
obvious thing to suspect.

### Findings

|---|---|---|---|---|
| 1 | Medium (P2) | `tests/unit/test_build_options_are_read.py:52` | **The guard matches one of the three ordinary ways to forward a compile definition, so two of them re-introduce a dead switch silently.** `_forwarded_definitions()` is `re.findall(r"add_compile_definitions\(\s*(EBLDR_[A-Z0-9_]+)\s*\)", text)` — the closing paren must follow the name immediately. **Demonstrated by editing `CMakeLists.txt` and running the module:** `add_compile_definitions(EBLDR_DEAD_A=1)` → **4 passed**; `target_compile_definitions(eboot_core PRIVATE EBLDR_DEAD_B)` → **4 passed**; the plain `add_compile_definitions(EBLDR_DEAD_C)` → **1 failed** ✅. The `=1` form is the more likely of the two to appear, because it is what someone writes when they want the macro to have a value, and `target_compile_definitions` is what they write when they want it scoped to one library. Neither is exotic. The guard is correct about the tree as it stands — `CMakeLists.txt` forwards exactly one `EBLDR_` definition, `EBLDR_VERIFY_STAGE1` at `:56`, and I confirmed there is no `target_compile_definitions` with an `EBLDR_` name anywhere — so this is entirely future-facing, which is also where the whole value of this guard lies. | Match the argument list rather than a single argument: collect every `EBLDR_[A-Z0-9_]+` token inside `add_compile_definitions(...)` **and** `target_compile_definitions(...)`, and strip anything after a `=` before comparing. Then add the two shapes above as negative cases, the way `test_no_option_offers_to_skip_verification_or_authentication` pins the two names. Secondary, same file: `test_documented_options_exist` checks documented ⊆ declared but not the reverse, so an option that exists and is documented nowhere passes. That direction is clean today — I enumerated it: declared is `EBLDR_{BOARD,BUILD_FUZZ,BUILD_TESTS,HARDENING,SANITIZE,VERIFY_STAGE1}`, `docs/book/book.md` lists all six and `README.md` lists five with `EBLDR_BOARD` covered in the prose above the table — so this is worth one more assertion, not a finding of its own. |
| 2 | Low (P3) | `tests/simulate_tests.py:169-181` | **This PR rewrites two assertions in a file that nothing executes, and the file has been failing for some time.** The change itself is exactly right — the old `test("Ed25519 signature requirement enabled by default", "EBLDR_REQUIRE_SIGNATURES" in cmake)` passed *because* the dead option was present, so it asserted the opposite of what its name claimed, and the replacement asserts the honest thing. But `grep -rn simulate_tests .github/ CMakeLists.txt tests/CMakeLists.txt Makefile*` is **empty**: no workflow, no CMake target, no Makefile runs it. I ran it here: **103/111, 8 failures, exit code 1** — including three version-drift failures (`build.yaml`, `CITATION.cff`, `CMakeLists.txt` version is 3.0.1) and five about CI jobs it believes are missing. None of that is visible to anyone. So the two corrected assertions are as unexecuted as the two vacuous ones they replace, and the PR body does not say so — a reader would reasonably take "the honest claim is the opposite one" as a check that now runs. The real guard here is `tests/unit/test_build_options_are_read.py`, which pytest does run, and the CHANGELOG correctly names that one. | Decide the file's fate in a follow-up rather than leaving it in between: either register it (`add_test` in `tests/CMakeLists.txt`, or a step in `ci.yml`) and fix or delete the eight stale assertions, or delete the file and move anything worth keeping into the pytest suite. Whichever way, say in this PR's body that the `simulate_tests.py` edit is a correctness repair to an unexecuted file, so the claim is not read as coverage. |

### What this PR gets right
The evidence is the strong part, and it is reproducible. The removal is justified by measurement
rather than by reading — a switch that produces identical objects is not a switch, and that is a
statement you can check, which I did. Choosing `FATAL_ERROR` over silently ignoring a stale
`-DEBLDR_REQUIRE_SIGNATURES=OFF` is the fail-closed reading: a caller passing that flag is asking for
something the bootloader does not do, and configuring as if the request had been granted is how the
original defect would have survived the cleanup. The `if(DEFINED ... AND NOT ...)` guard correctly
lets `=ON` through — asking for what you already get is not an error — and I confirmed `=OFF` and `=0`
both abort. `docs/book/book.md` losing `EBLDR_SECURE_BOOT`, `EBLDR_MULTICORE`, `EBLDR_RECOVERY` and
`EBLDR_BOOT_MENU` is the larger documentation repair: four options that **never existed** in
`CMakeLists.txt` were being presented to readers as configuration, and `test_documented_options_exist`
is what stops a fifth appearing. No stale references to either removed name survive outside the
three places that explain the removal.

Detached worktree at `.ai/autoreview/state/scratch/eBoot-133` on `88cb92e9`. The user's `eBoot`
touched. Nothing was committed or pushed. This review covers `a276016..88cb92e9` (one commit); the
rest of the bundle diff belongs to **#115**, which this branch is stacked on.

| Check | Result |
|---|---|
| `ctest --test-dir build/host --output-on-failure --no-tests=error` | **PASS — 31/31**, 10.98s |
| `pytest tests/ -q` | **PASS — 84 passed**, 1.79s |
| `pytest tests/unit/test_build_options_are_read.py -v` | **PASS — 4/4** |
| **The headline claim, reproduced.** Two `EBLDR_BOARD=none` **Release** builds, the second with `-DCMAKE_C_FLAGS="-DEBLDR_REQUIRE_SIGNATURES -DEBLDR_RECOVERY_AUTH"`, objects compared by SHA-256 | **41 identical, 0 differing** — exactly the author's "41 of 41". *Methodology note:* the same comparison on a **Debug** build shows 41/41 differing, because `-g` records the compiler command line in `DW_AT_producer`, so the flags themselves change the object. The claim is about generated code and holds; it needs a no-`-g` build to see. Worth a line in the body so the next person reproducing it does not conclude the opposite. |
| `grep -rn "EBLDR_REQUIRE_SIGNATURES\|EBLDR_RECOVERY_AUTH" --include=*.c --include=*.h` | **empty** — confirms no source ever read either macro |
| `cmake -DEBLDR_REQUIRE_SIGNATURES=OFF -DEBLDR_BOARD=none` | **`CMake Error at CMakeLists.txt:67`**, configure aborts with the intended message |
| `=ON` accepted, `=0` refused | Both as intended — `if(DEFINED ... AND NOT ...)` |
| **Guard blind-spot probes** (finding 1) | `add_compile_definitions(EBLDR_DEAD_A=1)` → **4 passed** ❌ · `target_compile_definitions(… EBLDR_DEAD_B)` → **4 passed** ❌ · `add_compile_definitions(EBLDR_DEAD_C)` → **1 failed** ✅. `CMakeLists.txt` restored; `git status --short` empty. |
| Declared vs documented options, enumerated through the module's own helpers | declared = `{BOARD, BUILD_FUZZ, BUILD_TESTS, HARDENING, SANITIZE, VERIFY_STAGE1}`; `book.md` lists all six; `README.md` lists five, `EBLDR_BOARD` in prose. **No phantom, no undocumented.** |
| `tests/simulate_tests.py` — is it run, and does it pass? | Run by **nothing**; `python3 tests/simulate_tests.py` → **103/111, exit 1**. `import re` is present at `:12`, so the new `re.search` calls resolve — I checked, because the diff adds the first `re` use in that block. |
| Leftover references to the removed options | Only the three sites that explain the removal (`CMakeLists.txt:60,65`, `README.md:74-75`). No workflow or script still passes either flag. |
| **`Build PDF` failure cause** — job log read, not inferred | `! LaTeX Error: Command \chead already defined.` at `l.1560 ...nel@ifstar {\@dblarg\@@@chead}{\@chead}}`, `exit code 43`. A template/`fancyhdr` collision in the preamble, before any of this PR's `book.md` content is typeset. **Not caused by this PR**; **#135** addresses it. |

### Architecture conformance
Conforms. §21: eBoot is Tier 1 Foundation; `CMakeLists.txt`, `README.md`, `docs/`, `tests/` are all
inside the owning repo, and §21.1 is not engaged. §5.1 dependency direction is untouched — nothing
about the module graph changes, only which macros reach the compiler. The clause this PR serves is
**§28, Status, Evidence and Claims Policy**, and it serves it unusually directly: a build option
presented as `ON` by default, with the meaning *"Require Ed25519 signatures for boot"*, is a
**Implemented**-grade claim (`code and functional tests`) made about something that had neither — the
definition reached the compiler and no code read it. §28's rule that a claim without evidence must be
labelled rather than asserted is what the README and `book.md` now comply with. §14.1's *"keep
implemented, experimental and planned security features explicitly separated"* — restated in
`.ai/security.md` as *"never let a planned feature read as implemented"* — is the same point from the
security side, and four phantom options in `book.md` were the worst instance of it in the tree.
§5.1's *"eBoot keeps the trusted computing base minimal and auditable"* is served by the
`FATAL_ERROR`: a TCB whose build accepts a flag asking it to skip verification, and silently does not
skip it, is not auditable in either direction.

### Proposed changes
```
In this PR, if convenient (both are small):
  test_build_options_are_read.py   match target_compile_definitions() too, and
                                   EBLDR_X=1 as well as EBLDR_X; the two shapes
                                   as negative cases                        (finding 1)
  PR body                          say the simulate_tests.py edit is a repair to a
                                   file nothing runs, and that the Debug-build object
                                   comparison needs -g off to reproduce      (finding 2)

Follow-up, not this PR:
  tests/simulate_tests.py          register it and fix its 8 stale assertions,
                                   or delete it and keep what is worth keeping
  test_build_options_are_read.py   assert declared subset documented, the other
                                   direction of test_documented_options_exist
```
This branch is stacked on **#115** and cannot land before it. `Build PDF` is red and will stay red
until **#135** merges; it is not a reason to hold this PR, but it does mean this PR cannot be judged
by an all-green tree.

**No fix PR opened.** `tests/unit/test_build_options_are_read.py` exists only on this branch — not on
would have nothing to patch. Finding 2 is Low and outside the autofix rule regardless.

### Not checked
- **The PDF build was NOT run locally.** No pandoc or LaTeX toolchain here. The cause above is read
  from the failing job's log; that **#135** fixes it is taken from its title and reviewed separately,
  not verified here.
- **Only the host `EBLDR_BOARD=none` build was compared.** The 41/41 result is Verified for that
  configuration. That the same holds for a cross-compiled board build is **Inferred** — it follows
  from no source file containing either macro, which I did verify, but I did not rebuild a board
  twice to measure it.
- **`EBLDR_HARDENING`, `EBLDR_SANITIZE`, `EBLDR_BUILD_FUZZ` were not audited for the same defect.**
  This PR's guard now covers them by construction for the `add_compile_definitions` form, and none of
  them is forwarded that way, but I did not check by hand whether each one actually changes the
  build. `EBLDR_SANITIZE` and `EBLDR_HARDENING` affect `CMAKE_C_FLAGS` rather than definitions, so the
  guard does not speak to them at all.
- **The #115 half of the diff was not reviewed here.** This review covers `a276016..88cb92e9` only.
- **Cross-compile, fuzz, ASan/UBSan, Valgrind, CodeQL, cppcheck/clang-tidy — NOT RUN locally.** The
  25 green checks were not re-run and their logs were not read; only the one red job's log was.


---
*Automated architecture review of `88cb92e9b262` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
Both taken. Restacked onto #115's `e152d8e` first (master `682d005`), then one commit on top; head is `25794f4`, range `e152d8e..25794f4`.

**Finding 1 (Medium) — the guard saw one of three shapes.** Reproduced your three probes first, same result (`=1` → 4 passed, `target_compile_definitions` → 4 passed, plain → 1 failed). `_forwarded_definitions()` now takes the whole argument list of `add_` or `target_compile_definitions(...)`, every `EBLDR_` token in it, `=value` stripped. Five shapes pinned as collected — including multi-line and a mixed list — and a second test appends each shape with a dead name to the *real* `CMakeLists.txt` text and requires the guard to fail, which is how a dead switch would actually return. All three of your probes now fail it. Your secondary note is taken too: declared ⊆ documented is asserted alongside documented ⊆ declared, with `EBLDR_BOARD` exempted for the reason you gave.

Negative control: old one-shape collector, new tests kept → **5 of 11 fail**.

**Finding 2 (Low) — `simulate_tests.py`.** Body now says it in the change list, in bold: a correctness repair to a file nothing executes, not coverage; the empty grep quoted; 8 stale failures by hand; the real guard is the pytest one. Its fate is a follow-up. The `-g` note is on the object-comparison line.

**On the red Build PDF:** pre-existing since 2026-04-25, as the body already said; the toolchain fix is #135 and this should land after it.

```
pytest tests/   93 passed
ctest           100% tests passed, 0 tests failed out of 31
```

(One correction visible in the history: the first push of this commit said `pytest 91` in its message, from the wrong head; amended to the measured 93 within a minute, `b5213df` → `25794f4`.) No approval existed at `88cb92e`. Could you re-review at `25794f4`.

Kartikey1306:
Restacked onto **#135** at `abf0522` — its Build PDF check just went green, proving the Eisvogel fix — so this PR's `Build PDF` can now pass instead of standing as a declared-prerequisite red. Head `f9fa26c`, range `abf0522..f9fa26c`, content unchanged (`range-diff` against `25794f4` empty). Merge order is now #115 → #135 → #133. At `f9fa26c`: ctest 31/31, pytest 93, 17 workflows parse.

PR #139, opened 2026-09-15
fix(recovery): INFO sent three bytes of stack to whoever asked, and the client printed them as the flash size

### LINKED ISSUE #140: RCVR_CMD_INFO sends three bytes of uninitialised stack to an unauthenticated caller; the client prints them as the flash size
`recovery_handle_info()` in `core/recovery.c` builds its response in a struct that is not `__attribute__((packed))` — unlike `rcvr_packet_t` fifteen lines above it — and never zeroes it. Only the six named members are assigned, then `eos_hal_uart_send(&info, sizeof(info))` transmits the whole object.

Measured, not argued:

```
sizeof(info) == 24   offsetof(flash_size) == 4
```

so bytes 1–3 are padding the code never writes. `RCVR_CMD_INFO` is not in `cmd_requires_auth()`, so any client on the UART can ask, repeatedly, before authenticating. With the struct poisoned to `0xA5` and filled exactly as the firmware fills it:

```
wire[0..8]: AA A5 A5 A5 00 00 10 00
```

Three bytes of whatever the previous call left on the stack, to an unauthenticated caller.

The repo's own client, `tools/uart_recovery.py`, reads `1 + 4*5 = 21` bytes and unpacks `'<IIIII'` from `response[1:21]` — it has always expected the **packed** layout. So it decodes `flash_size = 0x00A5A5A5` (the leaked bytes, printed as "Flash size"), every slot address one word off, and leaves three surplus bytes in the serial buffer to desynchronise whatever command follows.

`grep -rn "CMD_INFO\|handle_info" tests/` is empty. Nothing has ever exercised this command.

Raised by the automated review of #131 (finding 1, P1, pre-existing). Fix in #139: pack the struct and `memset` it — which makes the firmware send what the client always read, a repair rather than a format change — plus a test that drives `RCVR_CMD_INFO` through the `sim_ops` harness with the output buffer poisoned first.

Kartikey1306:
**Stacked on #131 (→ #127 → #115); review `00283e8..55a39d0` for this change alone** — three commits, seven files (corrected from `00283e8..2693259`, before that `00283e8..e14456f`; the second commit takes findings 1–5 of the 07:11Z review, the third takes finding 6 and in doing so replaces the finding-3 fix with a stronger one): `core/recovery.c`, `tools/uart_recovery.py`, `tests/unit/test_recovery.c`, `tests/unit/test_uart_recovery.py`. Merge order **#115 → #127 → #131 → this**; do not merge without #131, whose commit this carries.

Closes #140.

Takes finding 1 (**P1**, pre-existing) and finding 2 (P2) from the review of #131, in one change because both are in the same struct.

## The leak

`recovery_handle_info()` built its response in a struct that was **not packed** — unlike `rcvr_packet_t` fifteen lines above it — and **never zeroed**. On the wire that was 24 bytes: `ack`, three bytes of padding nothing wrote, then the five geometry fields. `RCVR_CMD_INFO` needs no authentication (`cmd_requires_auth()` does not list it), so any client on the UART could ask, repeatedly, and receive three bytes of whatever the previous call had left on the stack.

Reproduced with the struct poisoned to `0xA5` and filled exactly as the firmware fills it:

```
sizeof(info)=24  offsetof(flash_size)=4        (client expects 21, packed)
wire[0..8]: AA A5 A5 A5 00 00 10 00
client decodes flash_size=0x00A5A5A5 slot_a_addr=0x00001000   (actual 0x00100000 / 0x08010000)
```

`tools/uart_recovery.py` has always parsed the **packed** 21-byte layout, so it printed the leaked bytes as "Flash size", every slot address one word off, and left three surplus bytes in the serial buffer to desynchronise the next command. `grep -rn "CMD_INFO\|handle_info" tests/` was empty: nothing had ever exercised it.

## The fix

The struct is packed and `memset` before any field is written. That makes the firmware send what the client always read — a **repair, not a format change** — and means a future field cannot open a new hole.

One byte is appended: `caps`, `RCVR_CAP_RNG | RCVR_CAP_OTP`, derived from the ops table. This is finding 2: #131 makes authenticated recovery unreachable *by construction* on a board with no entropy source, and the event it logs is readable only over `RCVR_CMD_LOG`, which needs the authentication the missing entropy prevents. What an integrator saw was a bare `NACK` — indistinguishable from a wrong secret — after `0+1+2+4+8 = 15 s` of backoff. The unauthenticated INFO response answers that before the first AUTH is sent, and reveals nothing an attacker could not learn by trying.

The client reads 22 bytes, decodes `caps`, and says in words when a board has no authenticated recovery. **Correction (finding 2, 07:11Z review):** this paragraph originally said "21 is the old length". No firmware ever sent 21 bytes — 21 was only ever what the client *read*. The legacy length is **24**, and the first version of this client decoded it: `read(22)` of 24 bytes succeeds, so it printed the leaked padding as the flash size with a fabricated "no entropy" line and left two bytes to desync the next command — the PR's own defect, still shown against every unpatched device. `2693259` fixes that: after `read(22)` the client probes for surplus bytes (0.25 s timeout of its own), and any surplus means pre-fix firmware — it says so, drains, and returns `False`. A truncated read is still refused too.

## Tests, and what each is for

| test | pins |
|---|---|
| `test_info_sends_the_packed_layout_the_client_parses_and_nothing_else` | `out_buf` poisoned `0xA5` first; INFO via `sim_ops`; `out_len == 22`; every field at the client's offsets; the byte *after* the response still `0xA5` |
| `test_info_reports_a_board_without_an_entropy_source` | ops with `rng_get = NULL` → `caps` has RNG clear, OTP set |
| `test_info_decodes_the_packed_22_byte_response` | a 22-byte response decodes to the printed geometry and leaves the next byte **unread** |
| `test_info_names_a_board_with_no_entropy_source` | OTP-only `caps` → the message names the missing entropy |
| `test_info_refuses_the_24_byte_response_of_pre_fix_firmware` | the reviewer's reproduction byte for byte: `info()` is `False`, "pre-fix firmware"/"24 bytes" printed, `10601K` and `Entropy:` **not** printed, buffer empty, port timeout restored |
| `test_info_probe_restores_the_timeout_on_the_good_path` | the 0.25 s probe timeout is restored after a good response |
| `test_info_refuses_a_short_response` | a truncated (21-byte) read is refused — kept, docstring corrected |

## Negative controls

Each file reverted to #131's version with the tests kept, `touch` + rebuild both ways, recompilation confirmed in the build log:

```
core/recovery.c at #131         ->  fails at tests/unit/test_recovery.c:578: out_len == INFO_WIRE_LEN   (sends 24)
tools/uart_recovery.py at #131  ->  3 of 7 fail
```

For `2693259`:

```
tools/uart_recovery.py at e14456f, tests kept          ->  1 of 9 fails (the 24-byte case)
__attribute__((packed)) and the _MSC_VER pragma stripped ->  _Static_assert fires: "INFO response must be 22 bytes packed"
```

## Also in `2693259`

- **Finding 3, then finding 6 (`55a39d0`):** `2693259` mirrored the `_MSC_VER` pragma and asserted `sizeof == 22`. `55a39d0` removes the struct entirely: the response is built into `uint8_t[RCVR_INFO_LEN]` with an explicit little-endian store and the fill is checked to be exactly 22 bytes. That makes both findings true by construction — a byte buffer has no padding on any toolchain, and the words are little-endian regardless of the target — so the pragma and the assert from `2693259` are gone with the struct they guarded. `rcvr_packet_t`'s own packing is untouched.
- **Finding 4:** **ADR-021** — `INFO` stays unauthenticated deliberately (everything it discloses an attacker learns by trying; the integrator needs it before `AUTH`), `LOG` stays authenticated. T-404 records `caps`, narrows its mitigation to `LOG`, cites the ADR.
- **Finding 5:** `docs/architecture.md`'s command table gives the 22-byte layout field by field.

## At `55a39d0`

```
ctest            100% tests passed, 0 tests failed out of 31
test_recovery    9/9
pytest tests/    93 passed
```

`core/recovery.c` is CRLF and stays CRLF (545 → 574 lines). Attribution grep empty.

**NOT RUN:** a real UART session. The `sim_ops` harness carries the same bytes through `eos_hal_uart_send` that a board would; the client tests use a `FakeSerial` with the byte-exact response.


srpatcha (commented):
## Review — eBoot#139 "fix(recovery): INFO sent three bytes of stack to whoever asked, and the client printed them as the flash size"
head: e14456f72d3913f73686d9db3efff9ddf6df3d37  author: Kartikey1306  ci: pass

Reviewed the incremental commit only, as the PR body asks: `00283e8..e14456f`
(`core/recovery.c`, `tools/uart_recovery.py`, `tests/unit/test_recovery.c`,
`tests/unit/test_uart_recovery.py`). First look by this pipeline.

**Verdict:** The firmware-side fix is correct and well-tested — the struct is packed, zeroed before first write, and the new `test_recovery.c` case pins 22 bytes with the trailing byte still poisoned. The client-side half does not hold: against a device running today's unpatched firmware the new client still prints leaked padding as the flash size, now with a false "no entropy source" diagnostic attached, and the PR body states this case is refused when it is not.

### Findings

|---|---|---|---|---|
| 1 | High | `tools/uart_recovery.py:128-131` | `read(self.INFO_LEN)` against pre-fix firmware returns 22 of its 24 bytes, so `len(response) < self.INFO_LEN` is false and the response is decoded. Verified by running this PR's `uart_recovery.py` against a 24-byte unpacked response with the padding poisoned `0xA5`: `info()` returned `True` and printed `Flash size: 10601K` (`0x00A5A5A5`), `Slot A: 0x00001000`, `Slot B: 0x00000600`, `Entropy: NO -- authenticated recovery is unavailable on this board`, leaving `0600` in the buffer. That is the defect in the PR title, still present on every device not yet reflashed, plus a false hardware diagnostic and a two-byte desync of the next command. | Reject an over-long response, not only a short one. After `response = self.ser.read(self.INFO_LEN)`, do a further short read; if it returns anything, print that the device is running pre-fix firmware (24-byte unpacked INFO), drain the input buffer and `return False`. Add the 24-byte case to `test_uart_recovery.py`. |
| 2 | High | PR body, "The fix" | "It **refuses** a shorter response rather than decoding it: 21 is the old length, and a firmware still sending the unpacked 24 would otherwise be parsed one byte wrong forever." No firmware ever emitted 21 bytes — the old firmware emitted 24, which is the case the guard does not catch (finding 1). `test_info_refuses_a_short_response` therefore pins a length nothing produces while the real legacy length is untested. Per the review brief an unsupported claim is itself the finding. | Correct the claim, or make it true by implementing finding 1. Keep the 21-byte test and add a 24-byte one. |
| 3 | Medium | `core/recovery.c:265-277` | The packed attribute is applied only under `#if defined(__GNUC__) \|\| defined(__clang__)`. `rcvr_packet_t` at `core/recovery.c:53-68` — the struct this one is explicitly modelled on — also carries `#ifdef _MSC_VER / #pragma pack(push, 1) … pop`; the copy dropped it. On any toolchain outside that `#if`, `sizeof(info)` is 24 again, the padding leak returns, and nothing fails: there is no assertion pinning the size. "Cross-Platform (windows-latest)" does not catch it — `simulation-test.yml:88` only echoes a skip message. | Mirror the `_MSC_VER` pragma from `rcvr_packet_t`, and add `_Static_assert(sizeof(info) == 22, "INFO wire layout");` (or the repo's equivalent) so a toolchain that will not pack the struct breaks the build instead of shipping the leak. |
| 4 | Medium | `docs/threat_model.md:142` | T-404 records INFO as an information-disclosure surface whose planned mitigation is "Restrict INFO/LOG to authenticated sessions (Phase 3)". This PR adds a new field to that unauthenticated response and builds a diagnostic workflow on its being reachable *before* AUTH — the whole point of `caps`. The threat model is not updated, and the two are now in direct conflict: implementing T-404's mitigation would remove the capability this PR relies on. | Update T-404 in the same PR: record `caps` as disclosed, and either narrow the planned mitigation to LOG or state why INFO is now deliberately exempt. This is a design decision worth an ADR alongside ADR-020, not a silent row. |
| 5 | Low | `docs/architecture.md:99` | The command table describes INFO as "Flash layout and slot addresses". The response now also carries a capability bitfield, and its length is part of the contract between firmware and `tools/uart_recovery.py`. | Add `caps` and the 22-byte packed layout to the table, so the wire format has one written definition. |
| 6 | Low | `core/recovery.c:270-274` | The five `uint32_t` go on the wire in target byte order; `uart_recovery.py` unpacks `<IIIII`. Correct on every current eBoot target and pre-existing, but the struct is now the declared wire format rather than an accident, so the assumption should be written down rather than inferred. | One comment line stating the protocol is little-endian, or explicit `htole32` conversions. |

### Architecture conformance

Conforms. `core/recovery.c` is shared boot logic in eBoot's `core/` per `.ai/architect.md`'s eBoot layout, and `tools/uart_recovery.py` is host-side tooling — neither introduces a dependency pointing up a tier. Master design §8.1 requires a factory/recovery image strategy; hardening the recovery protocol's only unauthenticated diagnostic is squarely inside that. No change to the trusted computing base boundary: INFO is not in `cmd_requires_auth()` (`core/recovery.c:107-120`) before or after. eBoot is Tier 1 (§21) and the change stays in it.

Against §14.1 and `.ai/security.md`, the direction of the fix is right — the struct now fails closed on a field nobody wrote, and `memset` makes a future field safe by construction. The one security-relevant judgement call, widening an unauthenticated response, is argued in the PR body but contradicts a documented planned control (finding 4).

### Proposed changes

Smallest sequence that keeps things working:

1. `tools/uart_recovery.py`, in `info()`, after the 22-byte read:

       if self.ser.read(1):
           print("Failed to get device info: device sent more than "
                 f"{self.INFO_LEN} bytes — pre-fix firmware sends a 24-byte "
                 "unpacked INFO. Reflash before trusting this output.")
           try:
               self.ser.reset_input_buffer()
           except AttributeError:
               pass
           return False

   `FakeSerial` already supports `read`; the `reset_input_buffer` guard keeps the
   existing tests working unchanged.

2. `tests/unit/test_uart_recovery.py`: add `test_info_refuses_a_legacy_unpacked_response`
   feeding `bytes([0xAA]) + b"\xA5\xA5\xA5" + struct.pack("<IIIII", …)` and asserting
   `info() is False`. Negative control: the test fails against this PR's client.

3. `core/recovery.c`: wrap the struct in the same `#ifdef _MSC_VER` / `#pragma pack`
   pair `rcvr_packet_t` uses, and add a static assertion on `sizeof(info) == 22`.

4. `docs/threat_model.md` T-404 and `docs/architecture.md:99` updated in the same
   change, per the project rule that a change making existing documentation wrong
   is not finished.

Steps 1–2 are independent of 3–4 and are the ones that matter for anyone holding an
unpatched board.

### Not checked

- A real UART session against real hardware. The PR says so itself; so do I.
- I did not build or run `ctest` for this branch. The PR reports `100% tests passed, 0 of 31` and `91 passed` at `e14456f`, and CI is green on all 26 reported checks; I did not re-run either. **NOT RUN**, not verified.
- The negative controls described in the PR body (reverting each file to #131 and confirming the new tests fail) were not reproduced.
- Findings 1 and 2 were verified by executing this PR's `tools/uart_recovery.py` against a synthetic 24-byte legacy response outside the repo tree; the output quoted above is real. Finding 3 is read from the source, not demonstrated on MSVC — no MSVC toolchain was available and CI does not compile there.
- The rest of the stack (#115 → #127 → #131) was not re-reviewed here; only `00283e8..e14456f` was judged. The 31-file diff in the bundle is the cumulative stack, not this change.
- `mergeStateStatus` is `BLOCKED` — that is the stacked-merge-order requirement the PR describes, not a failing check, but I did not confirm which branch protection is holding it.


---
*Automated architecture review of `e14456f72d39` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

Kartikey1306:
All six taken, in two pushes. Head is `55a39d0`; every number below was measured there.

**1 (High) — the client decoded pre-fix firmware's 24 bytes.** You were right, and it is the worst kind of wrong: the PR's own defect, still shown by the PR's own client, against every device not yet reflashed, with a fabricated "no entropy" line on top. Reproduced first, exactly your way — `info()` returned `True`, printed `Flash size: 10601K`, left `0600` in the buffer. Now: after `read(22)` the client probes for surplus bytes with a 0.25 s timeout of its own (the port's 5 s is for flash work; two bytes at any baud arrive well inside a quarter second), and any surplus means pre-fix firmware — it names the length, says the bytes after ACK are uninitialised stack, drains, returns `False`, restores the timeout on every path. A short read rather than `in_waiting`, because on a real UART those two bytes may still be on the wire when `read(22)` returns.

**2 (High) — "21 is the old length."** Wrong, and the body now says so where it said it: no firmware ever sent 21; 21 was only ever what the client *read*. The legacy length is 24, and **your reproduction is now the test** — `test_info_refuses_the_24_byte_response_of_pre_fix_firmware`, `0xA5` padding, asserting `info()` is `False`, "pre-fix firmware" and "24 bytes" printed, `10601K` and `Entropy:` **not** printed, buffer empty, timeout restored. The 21-byte test is kept as the truncated-read case with its docstring corrected. My earlier 22-byte test planted a stray trailing byte and asserted it stayed unread — that premise belonged to the check-less client and is gone.

**3 (Medium) and 6 (Low), together.** `2693259` did what you asked for 3: the `_MSC_VER` pragma pair mirrored from `rcvr_packet_t`, plus `_Static_assert(sizeof(struct rcvr_info) == 22)` — and I confirmed the assert fires with both guards stripped. Then finding 6 made me look again: a wire format whose byte order depends on the target's is an assumption, not a contract, and the right shape for a declared wire format is bytes, not a struct. `55a39d0` builds the response into `uint8_t[RCVR_INFO_LEN]` with an explicit little-endian store and checks the fill is exactly 22. That makes 3 and 6 both true *by construction* — a byte buffer has no padding on any toolchain and no byte order of its own — so the pragma and the assert are gone with the struct they guarded. The serialiser was checked on a value whose bytes differ by order (`0x08010000 → 00 00 01 08`). If you would rather keep the belt with the braces I will put the length assert back, but there is no struct left to pack.

**4 (Medium) — T-404.** This one is a design decision and I have treated it as one: **ADR-021** (Proposed, deciders as ADR-020). `INFO` stays unauthenticated deliberately — everything it discloses an attacker learns by trying, and the integrator needs it *before* `AUTH` to diagnose a board with no entropy source — and `LOG` stays authenticated, because boot history is the material T-404 is right to protect. T-404 now records `caps` as disclosed, narrows the mitigation to `LOG`, cites the ADR, and its status reads "Partial (INFO decided; LOG planned)". The ADR index has the row.

**5 (Low) — `docs/architecture.md`.** The command table gives INFO's 22-byte layout field by field, byte order stated, with a pointer to the ADR.

## Negative controls

```
tools/uart_recovery.py at e14456f, tests kept              ->  1 of 9 fails   (the 24-byte case)
__attribute__((packed)) + _MSC_VER pragma stripped (2693259) ->  _Static_assert fires
   (moot at 55a39d0: no struct)
```

## At `55a39d0`

```
ctest            100% tests passed, 0 tests failed out of 31
test_recovery    9/9   (le32() decode of every field, unchanged by the serialiser)
pytest tests/    93 passed
```

`core/recovery.c` is CRLF and stays CRLF. Range in the body is `00283e8..55a39d0` with both corrections visible. Merge order #115 → #127 → #131 → #139. No approval existed at `e14456f`; could you re-review at `55a39d0`.

PR #142, opened 2026-09-15
fix(sign_image): --genkey writes the key the build reads, not a header nothing includes

### LINKED ISSUE #141: quickstart's key-generation path writes a header nothing reads; firmware built by the documented procedure keeps the RFC test key
`docs/quickstart.md` § Signing Firmware Images tells a developer:

```
# Generate an Ed25519 keypair (first time only).
# Writes keys/private.pem, keys/public.pem and keys/public_key.h.
python3 sign_image.py --genkey --output keys/
```

`keys/public_key.h` defines `static const uint8_t ebldr_default_pubkey[32]`. That symbol has **zero consumers**:

```
$ git grep -n ebldr_default_pubkey
core/image_verify.c:177:   * symbol directly. Two reasons: ebldr_default_pubkey was never
tools/sign_image.py:343:static const uint8_t ebldr_default_pubkey[32] = {{
tools/sign_image.py:381:static const uint8_t ebldr_default_pubkey[32] = {{
```

`core/image_verify.c`'s own comment records that it "was never defined anywhere, so this did not link", and the verifier was changed to read the anchor from the keystore. The keystore's production slot is `ebldr_production_key[]`, generated at configure time from `-DEBLDR_PRODUCTION_KEY=<64 hex>` (see `docs/key_lifecycle.md`, #122). Nothing includes `public_key.h`.

So a developer who follows the documented procedure generates a key, gets a header, builds, sees no error, and ships firmware whose trust anchor is still the RFC 8032 §7.1 TEST 1 public key — whose private half is printed in a public IETF document. `.ai/security.md`: a weaker posture must be *documented*, not silently pretended strong. Here the documentation asserts the strong posture.

`--extract-pubkey` writes the same dead header. Raised as finding 2 (High, P1) in the automated review of #116 at `ef5b70d`, which also checked that #122 does not close it: after the whole stack lands the repo would document two key paths, one of which does nothing.

Kartikey1306:
**Stacked on #122 (→ #116 → #115); review `18fab20..180a74f` for this change alone** — two commits, four files. (Corrected from `18fab20..0c7c8f5`: the second commit fixes the same class in `docs/key_lifecycle.md`, which showed `--key-out` and `--pub-header`, two flags the tool has never had, and adds a guard that reads every documented `sign_image.py` invocation and requires each flag to exist.) Merge order **#115 → #116 → #122 → this**; the quickstart text this rewrites references what #122 creates, which is why the review asked for it on top of #122.

Closes #141.

## The defect

`docs/quickstart.md` tells a developer that `sign_image.py --genkey` writes `keys/public_key.h`. It did — a header defining `static const uint8_t ebldr_default_pubkey[32]`. That symbol has **zero consumers**:

```
$ git grep -n ebldr_default_pubkey
core/image_verify.c:177:   * symbol directly. Two reasons: ebldr_default_pubkey was never
tools/sign_image.py:343:static const uint8_t ebldr_default_pubkey[32] = {{
tools/sign_image.py:381:static const uint8_t ebldr_default_pubkey[32] = {{
```

`core/image_verify.c`'s own comment records it "was never defined anywhere, so this did not link"; the verifier reads the anchor from the keystore, whose production slot is `ebldr_production_key[]`, generated at configure time from `-DEBLDR_PRODUCTION_KEY=<64 hex>`. So the documented procedure generated a key, wrote a header, **built without error, and shipped firmware whose trust anchor was still the RFC 8032 test key** — whose private half is printed in an IETF document. The review of #116 (finding 2, High) found it and checked that #122 does not close it: after the stack landed the repo would document two key paths, one of which does nothing.

## The fix

Both emitters — `--genkey` and `--extract-pubkey` — now write the one artefact the build consumes: `public_key.hex`, 64 lowercase hex characters, the exact value `EBLDR_PRODUCTION_KEY` takes. The tool prints the `cmake` invocation that uses it, points at `docs/key_lifecycle.md`, and says to compare the value against the `.pub` before storing it (the one control for the ~1/16 typo case #122 documents). **The header is not written under any name** — a second mechanism for the same thing is how this drifted in the first place.

`docs/quickstart.md`'s signing section now shows the configure step with `-DEBLDR_PRODUCTION_KEY=$(cat keys/public_key.hex)`, says what happens without it, and names the lifecycle document.

## Tests

| test | pins |
|---|---|
| `test_genkey_writes_the_hex_the_build_consumes_and_no_dead_header` | no `public_key.h`; `public_key.hex` is 64 hex chars equal to the raw key in `public.pem`; stdout names the flag |
| `test_genkey_output_passes_the_configure_time_key_check` | the written value passes `tools/check_production_key.py` — the same check the gate runs — so the documented path cannot produce a key the build then refuses |
| `test_extract_pubkey_writes_hex_too` | same hex, no dead symbol |
| `test_quickstart_documents_the_key_path_the_build_reads` | names `public_key.hex` and the flag; not the header |

## Negative controls, one file reverted at a time, tests kept

```
tools/sign_image.py at 18fab20   ->  3 of 20 fail
docs/quickstart.md at 18fab20    ->  1 of 1 fails
```

## End to end, the documented procedure at `0c7c8f5`

```
$ python3 tools/sign_image.py --genkey --output keys/
$ cmake -B build -DEBLDR_BOARD=stm32f4 -DCMAKE_BUILD_TYPE=Release -DEBLDR_PRODUCTION_KEY=$(cat keys/public_key.hex)
configure rc 0
build/generated/production_key.c carries exactly the generated key   (byte-compared)
```

```
pytest tests/   142 passed   (at 180a74f; 138 at 0c7c8f5)
```

`tools/sign_image.py` is CRLF and stays CRLF (445 → 427 lines). Two corrections visible in the branch history: the phantom-flag guard's first version could not fail (its pattern never read a continuation line — caught because the negative control reported 1 failure where 2 were claimed; fixed in the same commit before this body was updated), and the first push of this commit had flattened it to LF — a 445-line diff for a 68-line change — caught on the line-ending check and amended before this PR was opened.


Kartikey1306:
**Ready for review — this PR has never had one, and it is the tip of the eBoot stack.**

Head `180a74f`, opened 09-15, unchanged since. Review `18fab20..180a74f` — two commits, four files:

- `0c7c8f5` — `--genkey` and `--extract-pubkey` both write `public_key.hex`, the 64 hex characters the build actually reads as `EBLDR_PRODUCTION_KEY`, through one `write_public_key_hex()`. The old `public_key.h` defined `ebldr_default_pubkey[32]`, which `git grep` shows has no consumers anywhere in the tree — `core/image_verify.c`'s own comment records that it "was never defined anywhere, so this did not link". Dead header removed.
- `180a74f` — `docs/key_lifecycle.md` showed `--key-out` and `--pub-header`, two flags `sign_image.py` has never had. Rewritten, plus a guard that reads every documented `sign_image.py` invocation and requires each flag to exist, so the docs cannot drift from the parser again.

**Verified at `180a74f` just now**, gcc/Ninja/Debug with `-DEBLDR_BUILD_TESTS=ON`:

- `ctest --no-tests=error` — 32/32 passed
- `pytest tests/` — **142 passed**, 0 skipped, with `cryptography` installed so the signing suite genuinely runs rather than being skipped

**Not run here:** the ARM and release legs, and the production-key gate's `FATAL_ERROR` path, which belongs to #122 below this. CI is green on all legs at this head.

Merge order: **#115 → #116 → #122 → this.** The quickstart text this rewrites references what #122 creates, which is why the earlier review asked for it to sit on top of #122 rather than on master. Closes #141.

